PR #42, opened 2021-11-12
AnimControl refactored to be a promise

### LINKED ISSUE #10: Control animation through the promise result of the animate() method
Currently, the chart's [animate()](https://lib.vizzuhq.com/0.3.0/reference/classes/default.html#animate) method returns a promise. The starting animation can be controlled (pause/play/seek) through the chart's [animation](https://lib.vizzuhq.com/0.3.0/reference/classes/default.html#animation) property.

    let anim = chart.animate();
    chart.animation.pause();

This two entity should be integrated together. The desired usage would be:

    let anim = chart.animate();
    anim.pause();

Current implementation:
https://github.com/vizzuhq/vizzu-lib/blob/main/src/apps/weblib/js-api/vizzu.js
https://github.com/vizzuhq/vizzu-lib/blob/main/src/apps/weblib/js-api/animcontrol.js

dyuri:
Closes #10 

2 integration tests are failing, gonna investigate:
- `example`
- `operations/filter_tutorial_data/filter_off_anim`

### DIFF
--- a/example/example.js
+++ b/example/example.js
@@ -96,6 +96,6 @@ let slider = document.getElementById("myRange");
 slider.oninput = (e)=>
 {
 	let t = e.target.value;
-	chart.animation.pause();
-	chart.animation.seek(t/10 + '%');
+	anim.pause();
+	anim.seek(t/10 + '%');
 };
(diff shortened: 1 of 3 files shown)


dyuri:
ok, sorry they were broken here even on `main`, updated the hashes, now they seem to be OK

dyuri:
forgot method chaining, please wait

simzer (approved):

PR #134, opened 2022-12-04
Fixed: UnPivot.convert method not modifying the original data object

### LINKED ISSUE #131: unpivot do not modify original data

veghdev:
fixed #131


simzer:
the original data obj could contain filter which will be lost this way.

veghdev:
Thanks, I modified it.

simzer (approved):

PR #153, opened 2023-01-06
No fade transitions v2

simzer:


veghdev (approved):

docs/content/tutorial/5-group-stack.md
```diff
@@ -1,41 +1,72 @@
...
+			x: { attach: ['Genres'] },
+			color: { attach: ['Genres'] }
 		}
 	}
 })
 ```
 
-By detaching this newly added dimension from the y-axis and attaching it to the 
-x-axis, you get a grouped bar chart in a way that is easy to follow for the viewer.
+To change the grouping of the chart, you just need to put replace the recently
```
petervidos, line 21: In order to change the category via which the elements are grouped, just replace the dimension with another one on the same axis.

docs/content/tutorial/5-group-stack.md
```diff
@@ -1,41 +1,72 @@
...
+			color: { detach: ['Genres'], attach: ['Kinds'] }
 		}
 	}
 })
 ```
 
-To stack a grouped chart, you just have to do the same thing the other way 
-around: detach the dimension from the x-axis and attach it to the y-axis.
+When the partitioning of the underlying data to markers on the chart is changing
+due to setting different dimensions for the chart, there are multiple ways to
```
petervidos, line 36: due to adding different dimensions to the chart

docs/content/tutorial/5-group-stack.md
```diff
@@ -1,41 +1,72 @@
...
 		}
 	}
 })
 ```
 
-To stack a grouped chart, you just have to do the same thing the other way 
-around: detach the dimension from the x-axis and attach it to the y-axis.
+When the partitioning of the underlying data to markers on the chart is changing
+due to setting different dimensions for the chart, there are multiple ways to 
+transition between the starting and end state.
```
petervidos, line 37: transition between these states.

docs/content/tutorial/5-group-stack.md
```diff
@@ -1,41 +1,72 @@
...
 
-```javascript { "title": "...doing it the other way is how you stack your chart" }
+You can aggregate the markers to the common base of the two states, as shown
```
petervidos, line 39: By default, the markers are aggregated to the common base of the two states, as shown above.
You can change this setting and drill down instead:

docs/content/tutorial/5-group-stack.md
```diff
@@ -1,41 +1,72 @@
...
+	config: {
+		channels: {
+			x: { detach: ['Kinds'], attach: ['Genres'] },
+			color: { detach: ['Kinds'], attach: ['Genres'] }
+		}
+	}
+}, { regroupStrategy: 'drilldown' })
+```
+
+or simply fade between them:
```
petervidos, line 53: There is also the option to fade the chart between the states:

docs/content/tutorial/5-group-stack.md
```diff
@@ -1,41 +1,72 @@
...
+	config: {
+		channels: {
+			x: { detach: ['Genres'], attach: ['Kinds'] },
+			color: { detach: ['Genres'], attach: ['Kinds'] }
+		}
+	}
+}, { regroupStrategy: 'fade' })
+```
+
+To get a stacked chart, you just have to move the added dimension to the same channel
```
petervidos, line 66: To stack a grouped chart,

src/apps/weblib/js-api/vizzu.d.ts
```diff
@@ -601,6 +601,15 @@ interface GroupOptions
 	delay?: Duration;
 }
 
+/** Algorithm type for transition in case of the data grouped 
+	differently on the source and target chart (different set of categorical
+	dataseries used).
```
petervidos, line 606: /** Type of transition when the categorical series differ on the source and the target chart.

PR #177, opened 2023-04-24
dataset integration into the vizzu-lib

tczagany:


simzer:
unnecessary data2 folder still present in the source

PR #184, opened 2023-06-01
Fix g++11 compilation + tests

schaumb:
- implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20
- control reaches end of non-void function
- Order of evaluation of function arguments is unspecified
- either all initializer clauses should be designated or none of them should be 
  **I kept some initialization comments for non-strongly typed variables in style.cpp.**
  - https://stackoverflow.com/questions/67005837/how-to-combine-base-class-explicit-constructor-and-designated-initialization-in
  - https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2287r1.html)

### DIFF
--- a/src/chart/main/style.cpp
+++ b/src/chart/main/style.cpp
@@ -20,37 +20,37 @@ Font Chart::defaultFont{
 Chart Chart::def()
 {
 	return {
-		{
+	    Padding {
 			.paddingTop = Gfx::Length::Emphemeral(5.0/12.0),
 			.paddingRight = Gfx::Length::Emphemeral(5.0/12.0),
 			.paddingBottom = Gfx::Length::Emphemeral(10.0/12.0),
 			.paddingLeft = Gfx::Length::Emphemeral(5.0/12.0)
 		},
-		{
+		Box {
 			.backgroundColor = Gfx::Color::White(),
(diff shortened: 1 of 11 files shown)


src/chart/main/style.cpp
```diff
@@ -68,32 +68,32 @@ Chart Chart::def()
...
-				.borderOpacityMode =
+				/*.borderWidth = */ 1,
+				/*.borderOpacity = */ 1,
+				/*.borderOpacityMode = */
 					::Anim::Interpolated<Marker::BorderOpacityMode>
 					(Marker::BorderOpacityMode::premultiplied),
-				.fillOpacity = 1,
-				.guides = {
+				/*.fillOpacity = */ 1,
+				/*.guides =*/ Guide {
```
schaumb, line 77: unnecessary

src/chart/main/style.cpp
```diff
@@ -68,32 +68,32 @@ Chart Chart::def()
...
 					.color = Gfx::Color::Gray(0.91),
 					.lineWidth = 0.5
 				},
-				.label = {
-					{
-						{
-							{
+				MarkerLabel{
+	                OrientedLabel {
```
schaumb, line 82: space/tab. fix with clangtidy

PR #191, opened 2023-06-05
Correct affine transformation implementation.

simzer:


project/cmake/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ add_subdirectory (test)
 
 if(NOT EMSCRIPTEN)
 
-find_package(Qt5 QUIET)
+find_package(Qt5 COMPONENTS Widgets Gui Core QUIET)
```
schaumb, line 10: The component selection will not work if Qt is not installed (hard error).

src/base/geom/affinetransform.h
```diff
@@ -15,33 +15,25 @@ class Polygon;
 class AffineTransform
 {
 public:
-	Geom::Point offset;
-	double scale;
-	double rotate;
+	double m[2][3];
```
schaumb, line 18: use std::array

https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-avoid-c-arrays.html

src/base/geom/affinetransform.cpp
```diff
@@ -7,108 +7,92 @@
...
-
-AffineTransform AffineTransform::operator*(double scale) const
-{
-	return AffineTransform(offset, this->scale * scale);
-}
-
-AffineTransform AffineTransform::operator*(const AffineTransform &other) const
-{
-	return AffineTransform(other(offset), other(scale));
+	auto det = m[0][0]*m[1][1] - m[1][0]*m[0][1];
```
schaumb, line 35: check determinant is not 0

src/base/geom/affinetransform.cpp
```diff
@@ -7,108 +7,92 @@
...
 
-AffineTransform::AffineTransform(Rect from, Rect to)
-{
-	rotate = 0.0;
-	scale = std::max(to.size.x / from.size.x, to.size.y / from.size.y);
-	offset = to.pos - from.pos * scale;
-	to.pos = from.pos * scale + offset;
-	to.size = from.size * scale;
-}
+AffineTransform::AffineTransform(Point offset, double scale, double angle) :
```
schaumb, line 26: create static builder function as this constructor replacement.

src/base/geom/affinetransform.h
schaumb, line 18: It can be made as private member. Then it need `JScriptOutputCanvas::transform` to be a friend function.

src/base/geom/affinetransform.cpp
```diff
@@ -7,108 +7,92 @@
...
+	return AffineTransform(
+		m[1][1] / det, 
+		-m[0][1] / det, 
+		(m[0][1]*m[1][2] - m[0][2]*m[1][1]) / det,
+		-m[1][0] / det, 
+		m[0][0] / det,
+		(m[0][2]*m[1][0] - m[0][0]*m[1][2]) / det);
 }
 
 bool AffineTransform::operator==(const AffineTransform &other) const
```
schaumb, line 45: Make it default: https://en.cppreference.com/w/cpp/language/default_comparisons

src/apps/qutils/canvas.cpp
```diff
@@ -368,9 +368,11 @@ Geom::Size BaseCanvas::textBoundary(const std::string &text)
 }
 
 void BaseCanvas::transform(const Geom::AffineTransform &transform) {
-	painter.translate(transform.offset.x, transform.offset.y);
-	painter.rotate(180.0 * transform.rotate / M_PI);
-	painter.scale(transform.scale, transform.scale);
+	const auto &m = transform.m;
```
schaumb, line 371: use structure binding to simplify the code:
```
const auto& [m0, m1] = transform.m;
```

src/base/geom/affinetransform.cpp
```diff
@@ -7,108 +7,92 @@
...
+	return *this != AffineTransform();
 }
 
-Geom::Size AffineTransform::operator()(const Geom::Size &original) const
+AffineTransform AffineTransform::operator*(const AffineTransform &other) const
 {
-	return original * scale;
+	AffineTransform res;
+	const auto &o = other;
+	res.m[0][0] = m[0][0]*o.m[0][0] + m[0][1]*o.m[1][0];
```
schaumb, line 64: use constructor aggregate initialization

src/base/geom/line.h
```diff
@@ -30,6 +30,8 @@ struct Line
 
 	Geom::Line extend(double length) const {
 		auto p = end - begin;
+		auto d = p.abs();
+		if (d == 0) return *this;
 		auto m = length / p.abs();
```
schaumb, line 35: Use `d`:
```
auto m = length / d;
```

src/base/gui/widget.cpp
```diff
@@ -163,8 +163,8 @@ Geom::AffineTransform Widget::getSelfTransform() const
 Geom::AffineTransform Widget::getTransform() const
 {
 	Geom::AffineTransform res;
-	if (parent) res *= parent->getTransform();
-	res *= getSelfTransform();
+	if (parent) res = res * parent->getTransform();
```
schaumb, line 166: unnecessary `res *` because it is an identity matrix.

schaumb (approved):

src/base/geom/affinetransform.h
simzer: JScriptOutputCanvas is in the application logic, wouldn't reference here in this very low level generic class.

src/base/geom/affinetransform.cpp
simzer: ok

src/base/geom/affinetransform.cpp
simzer: why?

src/apps/qutils/canvas.cpp
simzer: ok

src/base/geom/line.h
simzer: ok

src/base/gui/widget.cpp
simzer: ok

project/cmake/CMakeLists.txt
simzer: I will revert this for now, but we should find a solution works for both cases, because previous solution fails to build the qtest project:
```
/home/simon/Desktop/vizzu/vizzu-lib/project/cmake/CMakeLists.txt(10):  find_package(Qt5 QUIET )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5ConfigVersion.cmake(2):  set(PACKAGE_VERSION 5.12.8 )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5ConfigVersion.cmake(4):  if(PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5ConfigVersion.cmake(6):  else()
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5ConfigVersion.cmake(7):  set(PACKAGE_VERSION_COMPATIBLE TRUE )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5ConfigVersion.cmake(8):  if(PACKAGE_FIND_VERSION STREQUAL PACKAGE_VERSION )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake(2):  if(CMAKE_VERSION VERSION_LESS 3.1.0 )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake(6):  if(NOT Qt5_FIND_COMPONENTS )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake(7):  set(Qt5_NOT_FOUND_MESSAGE The Qt5 package requires at least one component )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake(8):  set(Qt5_FOUND False )
/usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake(9):  return()
```

src/base/geom/affinetransform.h
schaumb: OK :)

src/base/geom/affinetransform.cpp
schaumb: Like this:
```cpp
const auto &o = other;
return {{
		{{
			m[0][0]*o.m[0][0] + m[0][1]*o.m[1][0],
			m[0][0]*o.m[0][1] + m[0][1]*o.m[1][1],
			m[0][0]*o.m[0][2] + m[0][1]*o.m[1][2] + m[0][2]
		}},
		{{
			m[1][0]*o.m[0][0] + m[1][1]*o.m[1][0],
			m[1][0]*o.m[0][1] + m[1][1]*o.m[1][1],
			m[1][0]*o.m[0][2] + m[1][1]*o.m[1][2] + m[1][2]
		}}
	}};
```

src/base/geom/affinetransform.h
simzer: I can make it private with a const ref  getter:
```cpp
const std::array<std::array, 3>, 2> getMatrix() const;
```

src/base/geom/affinetransform.cpp
schaumb: Not required. I just thought that the parameters are not simply as first seems, and it is less expressive than a `AffineTransformation::from_units(point, double, double)` function next to an `AffineTransformation::AffineTransformation(double, double, double...)` constructor

project/cmake/CMakeLists.txt
schaumb: I think the `module mode` can be the solution, but i'm not a qt cmake expert:
https://cmake.org/cmake/help/latest/command/find_package.html#search-modes

src/base/geom/affinetransform.cpp
schaumb: It has user defined constructor, so the aggregate initialization will not work. But it has a 6 double parametered constructor. That can be used.

project/cmake/CMakeLists.txt
simzer: I will check only the qt core comonent, hopefully it will work
```

find_package(Qt5Core QUIET)
if( Qt5Core_FOUND )
    add_subdirectory (qtest)
endif()
```

schaumb: Nice solution in last commit, I resolve this CR

src/base/geom/affinetransform.cpp
simzer: I have created a constructor accepting the array and used that.

schaumb (approved):
Looks good, thanks

PR #194, opened 2023-06-09
Axis label fixes

simzer:


src/base/math/interpolation.h
```diff
@@ -10,7 +10,7 @@ template <typename T>
 T interpolate(const T &op0, const T &op1, double factor)
 {
 	if (factor <= 0.0) return op0;
```
schaumb, line 12: If the `factor >= 1.0` branch need `T() * 0.0`, then here isn't it necessary the `T() * 0.0` ?

src/base/math/interpolation.h
```diff
@@ -10,7 +10,7 @@ template <typename T>
...
-	else if (factor >= 1.0) return op1;
+	else if (factor >= 1.0) return T() * 0.0 + op1 * 1.0;
```
schaumb, line 13: This must be a T(), and not op0?

simzer: This is needed because ::Anim::Interpolated contains both operand outside, and should later tell, which is which.
If we just pass op1, then Interploated will consider it as op0.
Anyway, this function should not depend on the internal working of ::Anim::Interpolated, I will solve this differently

src/base/math/interpolation.h
simzer: It is not needed, but I will revert this anyway, see the other comment.

src/base/anim/interpolated.h
```diff
@@ -195,13 +207,13 @@ class Interpolated
 	}
 
 	template <typename T = Type>
-	Type calculate() const
+	T calculate() const
 	{
 		if (this->count >= 1)
 		{
-			auto res = this->values[0].value * this->values[0].weight;
+			auto res = (T)this->values[0].value * this->values[0].weight;
```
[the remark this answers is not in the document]
simzer: fixed

schaumb (approved):

PR #201, opened 2023-06-12
Remove BubbleChart v2, rename v1 to Impl

schaumb:


simzer (approved):
BubbleChartBuilder does not need to be a template class anymore.

simzer:
@veghdev desktop version won't build with old v13 clang, please update to clang-16

PR #202, opened 2023-06-12
Rename Continuous and Discrete to Measure and Dimension. Fix Continou…

schaumb:
…s typo.


src/data/table/columninfo.cpp
```diff
@@ -96,17 +96,17 @@ ColumnInfo::ContiType ColumnInfo::getContiType() const
...
 
 const ColumnInfo::ValueIndexes &
-ColumnInfo::discreteValueIndexes() const
+ColumnInfo::dimensionValueIndexes() const
 {
 	return valueIndexes;
 }
 
-const ColumnInfo::Values &ColumnInfo::discreteValues() const
+const ColumnInfo::Values &ColumnInfo::dimensionValues() const
```
simzer, line 104: `dimensionValues` could be called `categories`. According to our nomenclature, Dimension series are containing categories, and Measure series are containing values.

simzer (dismissed):

simzer (approved):

PR #211, opened 2023-06-15
Remove TODOs from code

schaumb:
Make a ticket from TODOs or do it, but do not store inside the code.


simzer:
todos are removed: https://github.com/vizzuhq/vizzu-lib/pull/215

PR #212, opened 2023-06-15
Fix reinterpret casts, remove const casts

schaumb:
After these modifications, only safe (close to the C/JS interface) reinterpret_cast are permissible inside the code

### DIFF
--- a/src/apps/weblib/jsfunctionwrapper.h
+++ b/src/apps/weblib/jsfunctionwrapper.h
@@ -17,7 +17,7 @@ private:
 	    void *>...);
 
 public:
-	constexpr JsFunctionWrapper(JsFun fun) : wrapper{fun}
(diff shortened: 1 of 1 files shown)


src/apps/weblib/jsfunctionwrapper.h
```diff
@@ -0,0 +1,44 @@
...
+
+	operator std::function<R(Ts...)>() const &&
+	{
+		if (fun)
+			return {std::move(*this)};
+		else
+			return {};
+	}
+
+	R operator()(Ts &&...ts) const { return fun(&ts...); }
```
simzer, line 35: Don't we need to check if fun is nullptr here?

schaumb: I hide all members to a private wrapper object, where the operator() implemented. No check needed here, because it's callable only when fun is not null.

simzer (dismissed):

PR #218, opened 2023-06-19
Data records can be specified via objects beside arrays on JS API.

simzer:
@petervidos: please check the tutorial change
@veghdev: please check the JS changes

### DIFF
--- a/test/integration/test_cases/features/data_input/object_records.mjs
+++ b/test/integration/test_cases/features/data_input/object_records.mjs
@@ -1,3 +1,10 @@
+const normalRecord = { Dim0: "a", Dim1: "A", Meas: 1 };
+const missingMeasureRecord = { Dim0: "b", Dim1: "B" };
+const missingDimensionRecord = { Dim0: "c", Meas: 3 };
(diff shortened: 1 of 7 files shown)


test/integration/test_cases/features/data_input/object_records.mjs
```diff
@@ -0,0 +1,31 @@
...
+            name: "Dim1",
+            type: "dimension",
+        },
+        {
+            name: "Meas",
+            type: "measure",
+        }
+    ],
+    records: [
+        { Dim0: "a", Dim1: "A" },
```
veghdev, line 17: @simzer Please add a test case with non-existing series names too. (If I understand correctly, it won't cause any problems because in the `recordObjectToArray` function, the `forEach` loop goes through the previously existing `seriesList`, so any new series will be ignored.)

veghdev: ... and another without existing dimension, for example without Dim1.

simzer: done

veghdev (approved):

PR #221, opened 2023-06-20
Todos removed, todo checker made case-insensitive.

simzer:


simzer:
@veghdev the PR does not contain any effective code change. What could be the reason for the failing tests?

veghdev:
An exception occurred in 3 tests, for example [ ERROR    ] [ 325  ] [ SyntaxError: Invalid or unexpected token ] test_cases/web_content/cookbook/rendering/custom_linetype

If we didn't change anything, the change could be in the newest chrome/chromedriver, because for them we are using the latest.

simzer:
Ok, I'm on it, Thx

veghdev:
@simzer As I can see, the error was with this import:  import('https://esm.sh/tinycolor2'), I couldn't reproduce it either, and after restarting the CI, it also succeed. Maybe it was just a temporary error with the esm cdn.

veghdev (approved):

PR #230, opened 2023-06-23
Tidy

simzer:


src/base/gfx/draw/roundedrect.cpp
```diff
@@ -23,8 +23,8 @@ void RoundedRect::corner(const Geom::Point corner,
     const Geom::Point &dir0,
     const Geom::Point &dir1)
 {
-	const double kappa = 0.5522847498;
-	// (4.0/3.0)*tan(M_PI/(2.0*n)), n = 4
+	constexpr double tan_piPer8 = 0.41421356237;
```
simzer, line 26: Would be nice to have a constexpr tan()

schaumb: https://github.com/lakshayg/compile_time/blob/master/include/compile_time/detail/math/trig.inl 
?

simzer: Nice, I meant in the std lib :)

schaumb (approved):

PR #235, opened 2023-06-28
Remove all -1's as extremal elements. Replace members to optional

schaumb:


src/chart/options/channel.cpp
```diff
@@ -89,10 +89,10 @@ bool Channel::isSeriesUsed(const Data::SeriesIndex &index) const
 	    || (dimensionIds.includes(index));
 }
 
-int Channel::findPos(const Data::SeriesIndex &index) const
+std::optional<int> Channel::findPos(const Data::SeriesIndex &index) const
```
simzer, line 92: findPos in Channel and Channels are not used, can be removed.

src/chart/ui/events.cpp
```diff
@@ -30,7 +30,7 @@ std::string PointerEvent::dataToJson() const
 		coords = chart->getCoordSystem().getOriginal(position);
 	}
 	return "\"element\":\"" + elementUnder + "\""
-	     + ",\"pointerId\":" + std::to_string(pointerId)
+	     + ",\"pointerId\":" + std::to_string(pointerId.value_or(-1))
```
simzer, line 33: here would be better to put null to json instead of -1 in case of not set

simzer (approved):

src/chart/ui/events.cpp
schaumb: I replaced to Conv::toString(pointerId).

src/chart/options/channel.cpp
schaumb: ok

simzer (approved):

PR #238, opened 2023-06-29
Add cmake after build command which prints wasm size

schaumb:
- Add cmake after build command which prints wasm size
- Add flto flag
- Simplify emcc.txt


schaumb:
image

simzer (approved):

PR #239, opened 2023-06-29
Calling wasm-opt on cvizzu.wasm, also printing gzipped size.

simzer:


veghdev:
> @veghdev do we set emscripten environment variables using emsdk.sh in the CI build env?

@simzer In the container, the vizzu-devenv.deb package comes preinstalled. During the post-installation process, it invokes emsdk/emsdk_env.sh. However, some functionalities may differ when running as the root user. For instance, emcmake may not be available on the PATH and needs to be explicitly called using the complete path /root/emsdk/upstream/emscripten/emcmake (if I remember well)

simzer:
it seems that EMSDK env variable is not set.

simzer:
is emsdk_env.sh invoked or sourced? it should be sourced to be able to set the PATH and other envs

veghdev:
> is emsdk_env.sh invoked or sourced? it should be sourced to be able to set the PATH and other envs

https://github.com/vizzuhq/ppa/blob/master/src/vizzu-devenv/DEBIAN/postinst
echo "source \"$userhome/emsdk/emsdk_env.sh\"" >> $userhome/.bashrc

veghdev:
> So the build runs as root? Can we call cmake like this? `source "<path to emsdk>/emsdk_env.sh & emcmake cmake ....."`

modify: https://github.com/vizzuhq/vizzu-lib/blob/main/ci/cloudbuild/vizzu-devenv-prebuild-docker.yaml

veghdev (approved):

PR #240, opened 2023-06-29
Add exceptioninfo to error message

schaumb:


src/apps/weblib/js-api/vizzu.js
```diff
@@ -73,7 +73,8 @@ export default class Vizzu {
       } catch (e) {
         if (Number.isInteger(e)) {
           let address = parseInt(e, 10);
-          let cMessage = this.module._vizzu_errorMessage(address);
+          let type = new this.module["ExceptionInfo"](address).get_type();
```
simzer, line 76: no need for slower string lookup. this.module.ExceptionInfo should be suffice.

schaumb: True, thx

simzer (approved):

PR #243, opened 2023-06-30
Remove catch clauses

schaumb:
This will unify the way it works on Emscripten and desktop platforms.

### DIFF
--- a/src/chart/options/config.cpp
+++ b/src/chart/options/config.cpp
@@ -35,7 +35,8 @@ void Config::setParam(const std::string &path,
 		auto it = accessors.find(path);
 		if (it == accessors.end())
 			throw std::logic_error(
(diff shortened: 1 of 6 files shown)


src/chart/options/config.cpp
```diff
@@ -35,7 +35,8 @@ void Config::setParam(const std::string &path,
 		auto it = accessors.find(path);
 		if (it == accessors.end())
 			throw std::logic_error(
-			    "invalid config parameter: " + path);
+			    std::string(path) + "/" + value
```
simzer, line 38: Here we should put path only once into the error message

schaumb: path is std::string, so conversion is unnecesarry. I unified all error message to the following format:

path[/value]: message

simzer (approved):

PR #247, opened 2023-07-04
Moved web_content/animated into web_content/removed

veghdev:
Keep, until we measure test coverage.


simzer:
shouldn't we just move out the folder from web_content?

test/integration/test_cases/test_cases.json
```diff
@@ -1326,142 +1326,142 @@
                 "7f0cd74"
             ]
         },
-        "web_content/animated/composition_comparison_pie_coxcomb_column_2dis_2con": {
+        "web_content/removed/animated//composition_comparison_pie_coxcomb_column_2dis_2con": {
```
simzer, line 1329: there is an unnecessary slash in the path

simzer (approved):

veghdev:
> shouldn't we just move out the folder from web_content?

maybe. where web_content_removed/... ?

simzer:
it's ok for now, then we should tidy the test_case folder in the near future

schaumb (approved):

PR #253, opened 2023-07-04
Unify logging + Add tag and time

schaumb:
https://github.com/orgs/vizzuhq/projects/10/views/1?pane=issue&itemId=32331162

### DIFF
--- a/src/base/io/log.cpp
+++ b/src/base/io/log.cpp
@@ -1,5 +1,7 @@
 #include "log.h"
 
+#include <ctime>
+
 using namespace IO;
 
 namespace
--- a/src/base/io/log.cpp
+++ b/src/base/io/log.cpp
@@ -2,11 +2,32 @@
 
 using namespace IO;
 
-Log::LogFunc Log::logFunc;
+namespace
+{
+static bool enabled = true;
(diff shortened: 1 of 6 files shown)


src/base/io/log.cpp
```diff
@@ -1,12 +1,35 @@
...
+{
+static bool enabled = true;
+static Log::LogFunc logFunc = [](std::string const &s)
+{
+	puts(s.c_str());
+};
+
+}
+
+LogRecord::LogRecord() : content("[vizzu] [YYYY-mm-ddTHH:MM:SSZ] ")
```
simzer, line 17: I'm not sure if we still need the timestamp in the log message. I don't see timestamps usually in the console log.

simzer (approved):

src/base/io/log.cpp
schaumb: I added a new flag 'setTimestamp'

schaumb:
I created a new commit but it didn't showed here:
https://github.com/vizzuhq/vizzu-lib/commit/efcad31b722ba1c36ffa299911d962ed237662fd

I create a new PR

PR #269, opened 2023-07-10
Refactor eventHandlers

schaumb:
This PR does not include any changes related to the EventDispatcher refactor.


src/base/util/eventdispatcher.h
```diff
@@ -30,7 +30,7 @@ class EventDispatcher
 	typedef std::shared_ptr<Event> event_ptr;
 	typedef std::function<void(Params &)> handler_fn;
 	typedef std::map<std::string, event_ptr> event_map;
-	typedef std::list<std::pair<handler_id, handler_fn>> handler_list;
+	typedef std::list<std::pair<std::uint64_t, handler_fn>> handler_list;
```
simzer, line 33: Do we still need handler_id then?

test/qtest/chart.cpp
```diff
@@ -31,7 +31,8 @@ void TestChart::prepareData()
 	table.addColumn("Cat2", std::span(cat2));
 	table.addColumn("Val", std::span(val));
 
-	chart.getChart().getEventDispatcher()["pointeron"]->attach(
+	chart.getChart().getEventDispatcher().getEvent("pointeron")->attach(
+	    0,
```
simzer, line 35: It seems strange that we have to pass the id form outside on the c++ interface.

simzer (changes requested):
Please put a line into the changelog about the fixed bug.

src/base/util/eventdispatcher.h
schaumb: This is part of the EventDispatcher refactor. Out of scope of this pr.

schaumb:
This should be tested manually because no automatic tests exists.

simzer (approved):

PR #271, opened 2023-07-10
Orientation selector

schaumb:


simzer:
reason for some test fail: up to this point, orientation selection run after all individual config parameter change. This caused a dependence of the final orientation on the order of the x, y channel setting. This was an unwanted behavior.

schaumb:
If I only sort X and Y, only 27 tc fails.

src/chart/animator/planner.cpp
```diff
@@ -336,7 +336,7 @@ bool Planner::verticalBeforeHorizontal() const
 	const auto &srcOpt = source->getOptions();
 	const auto &trgOpt = target->getOptions();
 
-	if (srcOpt->horizontal != trgOpt->horizontal
+	if (*srcOpt->orientation.get() != *trgOpt->orientation.get()
```
simzer, line 339: Could be isHorizontal() as well.

simzer (approved):

simzer (changes requested):
Please add a note to the changelog

simzer (approved):

PR #272, opened 2023-07-10
Tolerance interval for mouse event position

schaumb:


src/chart/rendering/items/drawitem.cpp
```diff
@@ -171,8 +171,8 @@ Geom::ConvexQuad DrawItem::lineToQuad() const
 
 	auto wBeg = lineWidth[0] * coordSys.getRect().size.minSize();
 	auto wEnd = lineWidth[1] * coordSys.getRect().size.minSize();
-
-	return Geom::ConvexQuad::Isosceles(pBeg, pEnd, wBeg * 2, wEnd * 2);
+	return Geom::ConvexQuad::Isosceles(pBeg, pEnd,
+	    std::max(10.0, wBeg * 2), std::max(10.0, wEnd * 2));
```
simzer, line 175: We will possibly want to use this function in the rendering code part in the future. Would be better to turn 10.0 to a parameter here with a default 0.0 value and pass the 10.0 from the bounds() method, just to prevent later confusion.

simzer (dismissed):

schaumb:
I was able to trigger a bad_optional_access by clicking randomly.

simzer (approved):

simzer:
please open a bug report if it's not opened already.

PR #274, opened 2023-07-12
Release 0.8

veghdev:


docs/tutorial/aggregating_data.md
```diff
@@ -0,0 +1,177 @@
...
+    }
+})
+```
+
+Next to the default logic of sum, there are a handful of other aggregation
+logics that are available in `Vizzu`: `min`, `max`, `mean`, `count` and
+`distinct`. Let's go through them to see how they work.
+
+!!! note
+    Count and distinct aggregation logics relate to dimensions like `Genres` and
```
simzer, line 72: Count does not need neither measure nor dimension,

docs/tutorial/aggregating_data.md
```diff
@@ -0,0 +1,177 @@
...
+        channels: {
+            x: {
+                set: ['Genres']
+            }
+        }
+    }
+})
+```
+
+Next to the default logic of sum, there are a handful of other aggregation
```
simzer, line 67: algorithm would be better wording than logic.

simzer (approved):

PR #278, opened 2023-07-14
Integration test, remove duplicates, fix names, add geometry tests

veghdev:


schaumb (approved):
I checked only the 2 recommended commit

PR #280, opened 2023-07-17
Sort setters

schaumb:


test/integration/tests/style_tests.json
```diff
@@ -1258,202 +1258,202 @@
...
         "plot/xAxis/ticks/position/animated_outside-center": {
             "refs": [
-                "4e7fdb8"
+                "53fd4ec"
             ]
         },
         "plot/xAxis/ticks/position/animated_outside-inside": {
             "refs": [
-                "9cd88ee"
+                "53fd4ec"
```
schaumb, line 1441: Same hash for these tests?

veghdev: Background bars orientation changed, the ticks moved to the yAxis
Screenshot from 2023-07-17 11-50-48

simzer:
Sorting the keys in the setter object does not achieve the same effect as the orientation fix.
With sorting, every x channel will be set before y, however, every y should be set before x to have the same effect as the fix. 

see: https://github.com/vizzuhq/vizzu-lib/commit/3f6ae0e59ef3de91e4c9bad571035db066b9fc37

PR #286, opened 2023-07-23
Temporarily lock chrome version

veghdev:


schaumb (commented):
Can you modify it so that it's not temporary but can be set using an environment variable?

schaumb (approved):

PR #293, opened 2023-07-31
Add auto_json with tests

schaumb:


src/base/conv/auto_json.h
```diff
@@ -0,0 +1,126 @@
+#ifndef BASE_AUTO_JSON_H
+#define BASE_AUTO_JSON_H
+
+#include <initializer_list>
+#include <span>
```
schaumb, line 5: Not needed

src/base/conv/auto_json.h
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	template <class T>
+	inline auto operator()(T &&val,
+	    const std::initializer_list<std::string_view> &il)
+	    -> std::void_t<decltype(toString(std::forward<T>(val)))>
+	{
+		pre(il);
+		primitive(std::forward<T>(val));
+		curr = &il;
```
schaumb, line 70: Move to pre

src/chart/main/style.cpp
```diff
@@ -441,7 +441,7 @@ Chart Chart::def()
 			.tooltip = {
 				Font
 				{
-					.fontFamily = ::Anim::String("Roboto, sans-serif"),
+					.fontFamily = ::Anim::String(),
```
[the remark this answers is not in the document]
schaumb: This was set here too and not only Chart font because fontparent was empty. Reflection solved it.

simzer (approved):

PR #294, opened 2023-08-01
String conversions returning JSON renamed to toJSON().

simzer:


simzer:
They are use in events, related tests checked manually.

src/base/geom/line.h
```diff
@@ -58,14 +58,14 @@ struct Line
 		return {at(t0), at(t1)};
 	}
 
-	explicit operator std::string() const
+	std::string toJSON() const
```
schaumb, line 61: [[nodiscard]] for all struct which has toJSON const member function

schaumb (approved):

schaumb (approved):

src/base/geom/line.h
simzer: done

PR #310, opened 2023-08-17
Add clang-tidy additional checks

schaumb:


src/chart/speclayout/tablechart.h
```diff
@@ -27,10 +27,11 @@ void TableChart::setupVector(std::vector<Item> &items,
 		if (item.enabled) size++;
 
 	auto rowsize =
-	    static_cast<size_t>(singleColumn ? 1.0 : ceil(sqrt(size)));
+	    static_cast<ssize_t>(singleColumn ? 1.0 : ceil(sqrt(size)));
```
simzer, line 30: why ssize_t?

src/chart/options/channel.h
```diff
@@ -30,7 +30,7 @@ class Channel
 
 	static Channel makeChannel(Type id);
 
-	Channel();
+	// Channel() = default;
```
simzer, line 33: Why leave the deadcode?

src/chart/options/channels.cpp
```diff
@@ -3,14 +3,15 @@
...
 
-Channels::Channels()
-{
-	for (auto type = 0U; type < std::size(channels); type++)
-		channels[static_cast<ChannelId>(type)] =
-		    Channel::makeChannel(ChannelId(type));
-
-	reset();
-}
+Channels::Channels() :
```
simzer, line 6: What is the benefit of this pattern over a simple loop?

src/chart/generator/plot.cpp
```diff
@@ -76,23 +76,23 @@ bool Plot::MarkerInfoContent::operator==(
...
-	axises = other.axises;
-	guides = other.guides;
-	dimensionAxises = other.dimensionAxises;
-	anyAxisSet = other.anyAxisSet;
-	style = other.style;
-	keepAspectRatio = other.keepAspectRatio;
-	markersInfo = other.markersInfo;
-}
+    options(std::move(options)),
+	style(other.style),
```
simzer, line 87: indentation missmatch

src/chart/speclayout/tablechart.h
schaumb: because `std::ldiv` only works with signed integers, and this variable is not used as `size_t`, only as `ssize_t` (and `double`).

schaumb: note: size variable is only an integer, so it's square root must be smaller than ssize_t max always, so there are no over/underflow anywhere.

src/chart/options/channels.cpp
schaumb: Because this is an array initialization where the elements are not default constructible from now.

src/base/conv/parse.h
```diff
@@ -24,9 +24,8 @@ template <typename To> To parse(const std::string &string)
 		return To(string);
 	}
 	else if constexpr (std::is_same_v<To, bool>) {
-		return string == "true"  ? true
-		     : string == "false" ? false
-		                         : throw std::bad_cast();
+		auto s = string == "true";
```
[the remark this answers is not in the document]
schaumb:
```
src/base/conv/parse.h:28:8: error: Expression can be simplified [clion-simplify,-warnings-as-errors]
                     : string == "false" ? false
                     ^
src/base/conv/parse.h:29:28: error: Expression can be simplified [clion-simplify,-warnings-as-errors]
                                         : throw std::bad_cast();
```

Another solution:

```
		return string == "true" ||
		       (string != "false" && (throw std::bad_cast(), true));
```
Is it better?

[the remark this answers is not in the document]
simzer: Still much less expressive than the original IMHO. I wonder what simplification possibility clion-simplify "sees" here.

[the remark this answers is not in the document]
schaumb: Probably an another solution: 
```
switch (hash(string)) {
case hash("true"):
  return true;
case hash("false"):
  return false;
default:
  throw std::bad_cast();
}

```
(`hash` is already defined in `cinterface.cpp`)

simzer (approved):

PR #330, opened 2023-09-08
Manual calls vizzu.detach before reload an iframe

veghdev:


veghdev:
@simzer I cannot reproduce the error (Aborted(RangeError: WebAssembly.instantiate(): Out of memory: Cannot allocate Wasm memory for new instance)), but added detach call

simzer (approved):

PR #339, opened 2023-09-13
Multichart

schaumb:


src/apps/weblib/interface.js
```diff
@@ -2,131 +2,141 @@ mergeInto(LibraryManager.library, {
...
     window.open(UTF8ToString(url), '_blank')
   },
-  setCursor: function (cName) {
-    Module.render.canvas().style.cursor = UTF8ToString(cName)
+  setCursor: function (target, cName) {
+    Module.renders[target].canvas().style.cursor = UTF8ToString(cName)
   },
-  canvas_frameBegin: function () {
-    Module.render.frameBegin()
+  callLater: function (func, obj, later) {
```
dovicsin, line 8: What is the option where we use such a function?
Surely it is a good solution to fix the max time at the code level?
What is the problem if the value is set too high? If it's not a problem, then only a default time value is needed.

schaumb: At C++ code there is a event loop which was a ["busy event loop"](https://github.com/vizzuhq/vizzu-lib/pull/339/files/9b38b0facd5db6b535b4c638ed97eaf7e39252b8#diff-743f718a07efe57b55e20f794a9ccd982d0feeb7837e17954af66b266d62f4a3L400-L402) until now. 
This function arguments are C objects (c function pointer and the attached object for the callback), and the third represent the timeout when needed to call this c function. It [used in interface.cpp:443](https://github.com/vizzuhq/vizzu-lib/pull/339/files/9b38b0facd5db6b535b4c638ed97eaf7e39252b8#diff-1f7d8ec66b278043d960c437f1da31d968b015ff856cce662b90e1a22abc26d6R443), which is only [called at tooltip](https://github.com/vizzuhq/vizzu-lib/pull/339/files/9b38b0facd5db6b535b4c638ed97eaf7e39252b8#diff-9b6aa0ccd686a5b09e16657752159628089ddf88953437f009014cbb44e16fe4R207-R208) with default current time (now --> `0`) argument.

We fix the minimum time of the callback, not the maximum (`Math.max(20, x)` -> `x<20` => `20`, `x>20` => `x`). This will result that it will not spam the js thread.


When tooltip separated from the C++-code and transformed to a js-plugin, the whole callback event handling can be removed. So this is only a temporary improvement before deletion.

simzer: Egy helyen használjuk a tooltiphez a markerek trackelésénél. Igazából a paraméter nem is kellene, mert settimeout(..., 0)-val equivalens ahogy használjuk

dovicsin: Köszi  az infót.

dovicsin (approved):

src/base/gui/widget.h
```diff
@@ -15,18 +15,34 @@ class Widget
 public:
 	virtual ~Widget() = default;
 
-	virtual void onPointerDown(const PointerEvent &event) = 0;
-	virtual void onPointerUp(const PointerEvent &event) = 0;
-	virtual void onPointerLeave(const PointerEvent &event) = 0;
-	virtual void onPointerMove(const PointerEvent &event) = 0;
-	virtual void onWheel(double delta) = 0;
-	virtual void onKeyPress(const Key &key,
+	virtual void onPointerDown(
```
simzer, line 18: Canvas parameter is out of context on the mouse event handlers for an abstract widget.
Anyway, this whole class could be thrown out, since we don't have multiple widget types anymore.

src/chart/ui/chart.h
```diff
@@ -56,8 +71,9 @@ class ChartWidget : public GUI::Widget
 	bool unprocessedPointerLeave{};
 	std::optional<int64_t> trackedMarkerId;
 	std::optional<int64_t> reportedMarkerId;
+	bool needUpdate{true};
```
simzer, line 74: Signalling the need for an update is redundant in this class. this is what the doChange callback for.
needUpdate was part of the JS specific solution to assync connect the chart's onchange to the rendering call.

src/apps/weblib/interface.js
schaumb: I'll remove the Math.max

src/apps/weblib/objectregistry.h
```diff
@@ -10,33 +12,43 @@ namespace Vizzu
...
 
-	Handle reg(const std::shared_ptr<void> &ptr)
+	template <class T> Handle reg(std::shared_ptr<T> ptr)
 	{
-		Handle handle = ptr.get();
-		objects.emplace(handle, ptr);
-		return handle;
+		Handle res{ptr.get()};
+		{
+			auto lock = std::lock_guard{mutex};
```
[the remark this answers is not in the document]
schaumb: I think the "C-api" is like an API that can be used from multiple threads. I don't think to it that it is a JS API. 

But the emscripten supports the threading, so if someone wants to use it that way (workers) this will work.

src/base/gui/widget.h
schaumb: When we throw out tooltips and mouse events, this should be not a problem.
Currently, we need the canvas because one widget can be drawn to multiple canvases. And widget can store information about Canvas.

src/chart/ui/chart.h
schaumb: We call `Interface::update` periodically
```
      this._updateInterval = setInterval(() => {
        this.render.updateFrame()
      }, 25)
```
So this should be refactored too. I think this is currently not the scope of the PR.

src/apps/weblib/interface.js
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/339/commits/7644d6eff8f1ac6b991064e88389ad5642222198

src/apps/weblib/js-api/vizzu.js
```diff
@@ -64,10 +68,14 @@ export default class Vizzu {
     }
   }
 
-  _call(f) {
+  _call(f, addChartParam = true) {
```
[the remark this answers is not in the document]
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/339/commits/7644d6eff8f1ac6b991064e88389ad5642222198

src/chart/ui/chart.h
schaumb: I created a ticket from this discussion

simzer (approved):

PR #342, opened 2023-09-17
Update prettier version, run over test

veghdev:
Please review these commits:
19a89cf
6bf35e9
52e16ba

Commit db60e1f does not contain any manual modification, just runs prettier over the project.


schaumb (dismissed):
We are using tabs in C++ code. This commit makes inconsistent with the js code. One tabulator character is one byte, two space is two, so the js files will be bigger with spaces (in debug mode).

simzer:
2 space is quasi-standard in JS. On the other hand, if we reformat every JS file in the project, we should use this opportunity to replace all the double quotes with single quotes.

simzer (changes requested):
We should use this config:
```
{
  "$schema": "https://json.schemastore.org/prettierrc",
  "semi": false,
  "tabWidth": 2,
  "singleQuote": true,
  "printWidth": 100,
  "trailingComma": "none"
}
```

veghdev:
> We should use this config:
> 
> ```
> {
>   "$schema": "https://json.schemastore.org/prettierrc",
>   "semi": false,
>   "tabWidth": 2,
>   "singleQuote": true,
>   "printWidth": 100,
>   "trailingComma": "none"
> }
> ```

Replaced in 1ee9d7c, 2c2e20d and reformatted in cce0d69

simzer (approved):

PR #343, opened 2023-09-18
Add clang-tidy check: google-build-using-namespace

schaumb:
remove all `using namespace`


simzer (changes requested):
This PR introduced a lot of code noise. 
I think it would be better to put namespace directives to the cpp modules as well if we want to switch this check on.

simzer (approved):
ok

PR #350, opened 2023-09-20
JS API side event handling.

simzer:
generic js api event handling

### DIFF
--- a/src/apps/weblib/js-api/events.js
+++ b/src/apps/weblib/js-api/events.js
@@ -5,24 +5,30 @@ export default class Events {
     this.eventHandlers = new Map()
   }
 
+  _isJSEvent(eventName) {
+    return eventName.startsWith('api-')
+  }
+
   add(eventName, handler) {
     if (eventName !== '' + eventName) {
       throw new Error('first parameter should be string')
     }
 
     if (!this.eventHandlers.has(eventName)) {
-      const func = (param) => {
-        this._invoke(eventName, param)
+      let cfunc = null;
+      if (!this._isJSEvent(eventName)) {
+        const func = (param) => {
+          this._invoke(eventName, param)
+        }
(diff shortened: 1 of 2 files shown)


src/apps/weblib/js-api/events.js
```diff
@@ -45,24 +51,27 @@ export default class Events {
...
-
-        for (const handler of [...this.eventHandlers.get(eventName)[1]]) {
-          const eventParam = JSON.parse(jsparam)
+        let eventParam
+        if (!this._isJSEvent(eventName)) {
+          const jsparam = this.vizzu._fromCString(param)
+          eventParam = JSON.parse(jsparam)
           eventParam.preventDefault = () => {
             this.vizzu._call(this.module._event_preventDefault)()
           }
```
schaumb, line 77: Can you move this "c" eventParam creation to a helper function?

schaumb (changes requested):

src/apps/weblib/js-api/events.js
```diff
@@ -72,13 +81,20 @@ export default class Events {
...
-
+        } else {
+          eventParam = param
+          eventParam.preventDefault = () => {
+            canceled = true
+          }
+        }
+        for (const handler of [...this.eventHandlers.get(eventName)[1]]) {
           handler(eventParam)
```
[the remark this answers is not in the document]
simzer: This was intentional, because there is some use case, where we need to change data associated with the event. But it is true, that event objects should be immutable, so I will introduce a middleware pattern as well to support these use cases and I will leave the event objects immutable.

src/apps/weblib/js-api/events.js
simzer: yes

dovicsin (approved):

schaumb (approved):

PR #353, opened 2023-09-21
Fixed test run if reference lib is not compatible with the test case

veghdev:
Screenshot from 2023-09-21 17-17-42


schaumb (approved):
it is hard to read. 
Can you transform it to
- Format strings, replacement of string concatenations ('+'s). (Only if it's simpler to read).
- Break the `createTestCaseResultFailure` code down to smaller functions

PR #359, opened 2023-09-26
Use one filter if same

schaumb:


src/chart/animator/animation.cpp
```diff
@@ -65,14 +65,22 @@ void Animation::addKeyframe(const Gen::PlotPtr &next,
...
 	else if (strategy == RegroupStrategy::aggregate) {
-		auto &&andFilter = Data::Filter(
-		    [srcFilter = target->getOptions()->dataFilter,
-		        trgFilter = next->getOptions()->dataFilter](
-		        const Data::RowWrapper &row)
-		    {
-			    return srcFilter.match(row) && trgFilter.match(row);
-		    },
-		    0);
+		auto &srcFilter = target->getOptions()->dataFilter;
```
simzer, line 68: could we put this into Data::Filter::operator&&()

schaumb:
Review: The condition is not the final. We need to keep the dimensions until labellevel if only dimensions on the stackaxis.

src/chart/animator/animation.cpp
```diff
@@ -133,7 +127,7 @@ void Animation::addKeyframe(const Gen::PlotPtr &next,
 	addKeyframe(begin,
 	    next,
 	    options,
-	    strategy == RegroupStrategy::drilldown);
+	    strategy != RegroupStrategy::fade);
```
simzer, line 130: I suspect that this change should be done for the first addKeyframe in line 117

simzer (approved):

src/chart/animator/animation.cpp
schaumb: OK I'll do that in the next PR

PR #367, opened 2023-10-03
Fixes

schaumb:
- interploation between two version of the same chart even if duration explicitly set
- Anim options should be applied for the whole anim when intermediate keyframes added


src/chart/animator/options.h
```diff
@@ -25,7 +25,8 @@ enum class SectionId {
 	y,
 	x,
 	tooltip,
-	connection
+	connection,
+	nothing
```
simzer, line 29: If we add nothing here, then we will have a nothing option group in the anim options object, so the user can say this:
```
chart.animate({...}, { nothing: { delay: x, duration: y, easing: whatever })
```
I don't think that we need this on the API.

src/chart/animator/animation.cpp
```diff
@@ -110,23 +110,54 @@ void Animation::addKeyframe(const Gen::PlotPtr &next,
 	}
 
 	auto begin = target;
+
+	std::size_t real_animation{};
+
 	if (intermediate0) {
-		addKeyframe(target,
+		real_animation += strategy == RegroupStrategy::fade
```
simzer, line 117: This is duplicated, we calculate this here and then in addKeyframe. I think we should remove it from addKeyframe and pass this as an argument instead.

src/chart/animator/animation.cpp
```diff
@@ -110,23 +110,54 @@ void Animation::addKeyframe(const Gen::PlotPtr &next,
...
+	real_animation +=
+	    strategy == RegroupStrategy::fade
+	    || !begin->getOptions()->looksTheSame(*next->getOptions());
+
+	auto real_options = options;
+	if (auto &duration = real_options.all.duration;
+	    real_animation > 1 && duration) {
+		*duration =
+		    ::Anim::Duration{static_cast<double>(duration.value())
+		                     / static_cast<double>(real_animation)};
```
simzer, line 139: We should take all.delay parameter into consideration as well.
Maybe it would be better solution to have an Animation::reTime() function here similar to Planner::reTime() instead of precalculating the durations before setting the keyframes.

simzer (changes requested):
Please leave a comment in Changelog.

src/chart/animator/options.h
schaumb: OK, I'll find another way.

src/chart/animator/animation.cpp
schaumb: Should we "scale" the pre-generated keyframes based on inner default animation?

schaumb: I solve it with pre-calculation. Is it good, or should I rewrite it?

test/integration/tests/fixes/38072036.mjs
```diff
@@ -0,0 +1,34 @@
...
+    chart.animate(
+      {
+        config: {
+          x: 'Foo',
+          y: ['Foo', 'Bar']
+        }
+      },
+      1
+    ),
+  (chart) => chart.animate({}, 2),
```
simzer, line 21: Unfortunately the runtime of the animation won't be checked here, since the test runner just pauses the anim and seeks into it in different percents.

simzer (approved):

test/integration/tests/fixes/38072036.mjs
simzer: @veghdev if we would use duration in seek() instead of percent, then we could detect animation speed related changes.

simzer: created: https://github.com/orgs/vizzuhq/projects/10/views/1?pane=issue&itemId=40614195

PR #378, opened 2023-10-06
Add docs build to CI

veghdev:


docs/assets/javascripts/mdchart.js
```diff
@@ -1,9 +1,7 @@
 class MdChart {
   constructor(data, vizzu, id) {
     this.data = data
-    this.vizzuLoaded = import(vizzu).then((vizzuUrl) => {
-      return import(vizzuUrl.default)
-    })
+    this.vizzuLoaded = import('../dist/vizzu.min.js')
```
simzer, line 4: can be static import now, no need to dynamically import it

simzer (approved):

docs/assets/javascripts/mdchart.js
veghdev: fixed

simzer (approved):

PR #386, opened 2023-10-12
Fix layouts

schaumb:


src/chart/main/style.cpp
```diff
@@ -415,9 +415,9 @@ Chart Chart::def()
 			.title = {
 				Padding
 				{
-					.paddingTop = Gfx::Length::Emphemeral(15.0 / 26.0),
+					.paddingTop = Gfx::Length::Emphemeral(15.0 / 26.0 - 0.192307679),
```
schaumb, line 418: This constant is~5/26

simzer (approved):

schaumb:
I missed the changelog note. I'll commit with another bug

PR #393, opened 2023-10-18
Fixed: should not change lignthess min/max order if max<min.

simzer:


src/base/math/range.h
```diff
@@ -17,6 +17,14 @@ template <typename T> struct Range
 		double shift;
 	};
 
+	static Range<T> Raw(const T &min, const T &max)
```
schaumb, line 20: Is the 2 parameter constructor used anywhere? If not, it can be used as not min-max construction. If used I think the min-max construction should be a factory function and not the raw

schaumb (changes requested):

src/base/math/range.h
simzer: It is used. Please provide an explanation for your opinion why do you think the current constructor should be the factory.

schaumb: I think this range class can be aggregate type. (aggregate: no user-defined constructor exists, and all class members should be public.) That will make `consteval static auto members()` function unnecessary.

I don't want you to make this type of aggregate, but the long-term goal is this I think.

simzer: This is an exception use-case where we need an irregular range. All other uses are needed to be "normalized"

schaumb: Another problem is that the 2-parameter constructor assumes an invariant that min should be lower than max. This invariant is not true, and probably that is why it created this factory function

schaumb: OK, then no changes required

schaumb (approved):

src/base/math/range.h
simzer: What about something like this?
```c++
template <typename T, bool positive = true> struct Range
{
  	template <typename = std::enable_if_t<positive>>
	Range(const T &x, const T &y) :
	    min(std::min(x, y)),
	    max(std::max(x, y))
	{}
```

simzer: i agree. this PR is a quick fix on the release candidate.

PR #399, opened 2023-10-25
Axis fixes

schaumb:


src/chart/options/options.h
```diff
@@ -29,14 +29,22 @@ namespace Vizzu::Gen
 class Options
 {
 public:
+	enum class LegendId { color, lightness, size };
```
simzer, line 32: Can we add some protection to ensure that the values are the same as in ChannelId?

simzer (changes requested):

src/chart/options/options.h
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/399/commits/040dd47d42ee003b05af4860699a3046eb3e01f5

src/chart/options/options.h
```diff
@@ -29,14 +29,35 @@ namespace Vizzu::Gen
 class Options
 {
 public:
+	enum class LegendId : std::underlying_type_t<ChannelId> {
+		color,
+		lightness,
+		size
+	};
+
+	static_assert(std::ranges::all_of(Refl::enum_names<LegendId>,
```
simzer, line 38: Why don't we simply set the enumerator values using the corresponding ChannelId values (`color = ChanneId::color`) instead of checking if they match? That way it would work out of box even if the ChannelId enum order changed instead of resulting in a compiler error.

src/chart/generator/axis.h
```diff
@@ -123,8 +134,7 @@ struct DimensionAxis
 	};
 	using Values = std::multimap<Data::MultiDim::SliceIndex, Item>;
 
-	Math::FuzzyBool enabled{false};
-	::Anim::String title;
+	bool enabled{false};
```
simzer, line 137: Why this shouldn't be an interpolatable bool?

schaumb: Currently, the variable is not used as interpolated. The enabled is set on interpolation manually, and not interpolated.
https://github.com/vizzuhq/vizzu-lib/blob/main/src/chart/generator/axis.cpp#L124

src/chart/options/options.h
schaumb: That would be no check that these have to be the same name. It can set the values with your recommendation, but this assert checks the names too.

schaumb: After the transformation, this is the code. I think this is unnecessarily complicated.
```cpp
	enum class LegendId : std::underlying_type_t<ChannelId> {
		color = static_cast<std::underlying_type_t<ChannelId>>(
		    ChannelId::color),
		lightness = static_cast<std::underlying_type_t<ChannelId>>(
		    ChannelId::lightness),
		size = static_cast<std::underlying_type_t<ChannelId>>(
		    ChannelId::size)
	};
```

simzer: can we have both?

simzer: it has the benefit of working instead of failing on channelid change. 
but I agree, this is much less readable than it could be without the strict linter config we use. This is a tradeof which has been made already all over the code, not just here.

simzer:
Please write note to the Changelog

src/chart/options/options.h
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/399/commits/4c387d7f169a9fd186c31ac1d3ed59b22493aabe

simzer (approved):

PR #408, opened 2023-10-28
Rollup mangle private properties in minify

simzer:
5 kb (8%) size reduction in vizzu.min.js


schaumb (commented):
Looks good. I approve it.

veghdev (approved):

PR #417, opened 2023-11-08
Fixed unhandled exception on canvas css height between 0..1

simzer:


brsd (approved):

schaumb (commented):
Please add change log

simzer:
> Please add change log

I did not add this to the changelog, because this bug was not present in released versions.

PR #418, opened 2023-11-08
fixed crash on mousemove with inconsistent renderedchart state.

simzer:


src/chart/rendering/renderedchart.h
```diff
@@ -59,9 +59,10 @@ class RenderedChart
 public:
 	RenderedChart() = default;
 	explicit RenderedChart(const CoordinateSystem &coordinateSystem,
-	    const Gen::Plot *plot = nullptr) :
+	    std::shared_ptr<const Gen::Plot> plot =
+	        std::shared_ptr<const Gen::Plot>()) :
 	    coordinateSystem(coordinateSystem),
-	    plot(plot)
+	    plot(std::move(plot))
```
schaumb, line 65: RenderedChart 2 parameter constructor only called in chart.cpp:131, and only with exactly 2 arguments. 
The 2nd argument default parameter is unnecessary here.

(It should be an assertion here that this argument cannot be 0, but it's not the scope of this bug)
(the default constructor is called only at the `Vizzu::Chart` implicit default constructor. This should not create an invalid rendered chart, so the default constructor should probably be removed, not scope of this bug)

schaumb (approved):

PR #426, opened 2023-11-14
Tutorial basics

nandordevai:


docs/tutorial/stacking_explanation.md
```diff
@@ -0,0 +1,13 @@
+# Stacking explanation
+
+In the following animation you can see an example how selecting different
+measures and dimensions cause different stacking behaviour in the chart.
+
+First we set the x channel to a dimension and an arbitrary constant value to
```
simzer, line 6: I would point out that the first phase shows that dimensions slice the axis to equal sections.

docs/tutorial/stacking_explanation.md
```diff
@@ -0,0 +1,13 @@
...
+illustrate stacking. Then by changing the constant to a measure the markers’
+width change according to the values. Finally when the dimension is removed the
+bars get ordered by the measure.
```
simzer, line 9: What happens under the hood is that if no dimension present, then the markers measured from the origo.

simzer:
I have reviewed and approved the changes in the docs folder.
@veghdev if the changes in tools are ok as well, than please set the PR to approved.

PR #431, opened 2023-11-20
Animation parameters

schaumb:


project/cmake/weblib/emcc.txt
```diff
@@ -47,7 +47,8 @@ set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} \
 '_chart_relToCanvasCoords',\
 '_chart_canvasToRelCoords',\
 '_chart_setKeyframe',\
-'_anim_control',\
+'_anim_control_setValue',\
```
simzer, line 50: anim_setValue and anim_control_setValue could be the same

src/base/anim/control.cpp
```diff
@@ -150,15 +143,48 @@ void Control::finish(bool preRun)
...
-	         && playState != PlayState::running) {
+	         && options.playState != PlayState::running) {
 		if (!finished && onFinish) {
 			onFinish(true);
 			finished = true;
 		}
 	}
 }
 
+void Control::setValue(std::string_view path,
```
simzer, line 158: This code was on the interface because reflecting method/params of a class shouldn't be the responsibility of the actual class.

src/base/anim/control.cpp
```diff
@@ -150,15 +143,48 @@ void Control::finish(bool preRun)
...
+    const std::string &value)
+{
+	if (path == "seek") { seek(value); }
```
simzer, line 161: it would be better, if we could just set 'position' with duration and percent.

simzer (approved):

src/base/anim/control.cpp
schaumb: OK, I moved back to the interface.

src/base/anim/control.cpp
schaumb: I rewrite to position and set "seek" ts function

simzer (approved):

PR #446, opened 2023-11-30
Move Logo boundaries to layout + Create DrawChart

schaumb:


src/chart/rendering/drawingcontext.h
```diff
@@ -20,31 +20,50 @@ class DrawingContext
...
 	const CoordinateSystem &coordSys;
 	Gfx::ICanvas &canvas;
 	Painter &painter;
-	const Gen::Options &options;
 	const Styles::Chart &rootStyle;
 	const Events &rootEvents;
 	const Layout &layout;
 	RenderedChart &renderedChart;
+
+	const Gen::Options &getOptions() { return *plot->getOptions(); }
```
simzer, line 66: could be const?

src/chart/rendering/drawingcontext.h
```diff
@@ -20,31 +20,50 @@ class DrawingContext
...
 	    canvas(canvas),
 	    painter(*static_cast<Painter *>(canvas.getPainter())),
-	    options(*plot.getOptions()),
-	    rootStyle(plot.getStyle()),
+	    rootStyle(styles),
 	    rootEvents(events),
 	    layout(layout),
 	    renderedChart(renderedChart)
 	{
+		if (plot) {
```
simzer, line 35: Why is this better to be inlined here than having it in the previous place, in a dedicated method getCoordSystem()?

src/chart/rendering/drawingcontext.h
schaumb: It will be const in a next 2-3 PR

src/chart/main/layout.h
```diff
@@ -22,12 +22,14 @@ class Layout
...
+	Geom::Rect logo;
 
 	void setBoundary(const Geom::Rect &boundary,
-	    const Gen::Plot &plot,
+	    const Styles::Chart &style,
+	    const Gen::PlotOptionsPtr &options,
 	    Gfx::ICanvas &info);
 
-	void setBoundary(const Geom::Rect &boundary, Gfx::ICanvas &info);
+	void setLogoBoundary(const Styles::Logo &logoStyle);
```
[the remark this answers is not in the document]
schaumb: OK, noted.

src/chart/rendering/drawingcontext.h
schaumb: because there was no cached coordSystem, and sometimes it was a faulty usage (eg: when coordinate transformation was queried from the API).
In a later version, this will be moved back to Chart::draw, and `CoordinateSystem` default arguments will be filled in the constructor.
kép

PR #449, opened 2023-12-04
HighResolution

schaumb:


simzer (changes requested):
These Low/High presets are arbitrary. They were optimized for the QT canvas. We should not hard-code these into the lib. Instead we need the PathSampler::dMax, PathSampler::hMax parameters on the API.

schaumb:
I created another [PR](https://github.com/vizzuhq/vizzu-lib/pull/550).

PR #451, opened 2023-12-05
Fix unit + add prev changelog

### LINKED ISSUE #450: Add extra space between number and unit in the scale label, just like on marker label
image

schaumb:
Closes #450


src/base/text/smartstring.cpp
```diff
@@ -77,14 +80,15 @@ std::string SmartString::fromNumber(double value,
...
+	     + (unit.empty() || unit == "%" ? unit : " " + unit);
 }
 
-std::string SmartString::escape(const std::string &str, char specChar)
+std::string SmartString::escape(const std::string &str)
 {
 	std::string result;
 	for (const auto &ch : str) {
-		if ((ch == '\\') || specChar == ch) result.push_back('\\');
+		if ((ch == '\\') || '"' == ch) result.push_back('\\');
```
simzer, line 91: `ch == '"'` generally considered more readable.

src/base/text/smartstring.cpp
```diff
@@ -51,7 +51,8 @@ std::vector<std::string> SmartString::split(const std::string &str,
 std::string SmartString::fromNumber(double value,
     NumberFormat format,
     size_t maxFractionDigits,
-    const NumberScale &numberScale)
+    const NumberScale &numberScale,
+    const std::string &unit)
```
simzer, line 55: If it has a unit as well, then it is no longer a number->string converter, but rather a physical value -> string formatter. The name should reflect this change.

src/base/text/smartstring.cpp
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/451/commits/cac6ec1fa17fc958abffcb8649697feef37990ce

simzer (approved):

PR #457, opened 2023-12-17
Type-fest dependency fix

simzer:


veghdev:
regenerated the lock file because the dependencies have changed

veghdev (approved):

PR #465, opened 2024-01-08
fix json escape

schaumb:


simzer (dismissed):

simzer (commented):
Shouldn't we escape the object keys as well?

simzer (changes requested):
Changing back from approved because open question.

schaumb:
> Shouldn't we escape the object keys as well?

There is a [template argument](https://github.com/vizzuhq/vizzu-lib/blob/f4508fc41ac8acf455e4f1c76132db2593d4cc91/src/base/conv/auto_json.h#L292) for that. By default, it's not because the reflected member names are used as keys, which cannot contain escapable characters.

But there is a bug [here](https://github.com/vizzuhq/vizzu-lib/pull/463/files#diff-ceeaf2623a7d23f860fbf1d35cb31db4355c0652d82d69f5f0076821bd605620R298).

simzer (approved):

PR #484, opened 2024-01-26
Fix for 480

### LINKED ISSUE #480: Invisible legend modifies plot area size
Legend takes up space on the left side even when it's empty.
[mystory_issue.zip](https://github.com/vizzuhq/vizzu-lib/files/14051443/mystory_issue.zip)

schaumb:
closes #480 

(probably this check is not in the right place.)


src/chart/options/options.cpp
```diff
@@ -233,10 +233,13 @@ uint64_t Options::generateMarkerInfoId()
 
 void Options::setAutoParameters()
 {
-	if (legend.get().isAuto()) {
-		LegendType tmp = legend.get();
-		tmp.setAuto(getAutoLegend());
-		legend = tmp;
+	if (auto [leg, w] = legend.get(0); leg.isAuto()) {
```
simzer, line 236: I think we shouldn't have interpolated parameters here, so it would be safe to just use get(), and not checking the weight.

simzer:
I'm not sure if we would like to have this behaviour. If the use explicitly sets a legend, they might want to have the space even if there is no legend. Usecase: there are 3 chart in a data story, 1st and 3rd have legend, but not for the 2nd. In this case the user might not want the chart to get wider then narrower.

schaumb:
If the user wants to explicitly set a legend without drawing, then we need an option for that (an empty channel).

PR #495, opened 2024-02-21
Fixed stacked empty aggregated (min/max/mean) values

schaumb:


src/chart/generator/marker.cpp
```diff
@@ -197,7 +197,7 @@ double Marker::getValueForChannel(const Channels &channels,
 
 	auto measure = channel.measureId;
 
-	double value{};
+	double value = channel.defaultValue;
```
simzer, line 200: This default value is a channel-wise value with the purpose of providing a value for the channel if no data series is attached to it to result in the desired chart. So for example x and y default value is 1. We should not use this value for markers not having valid aggregators.

I think we should use 0 for all aggregators if they are empty.

simzer: Here because we use 1 height for the missing markers, the valid markers are floating: https://jsfiddle.net/simzer/thbenzLv/11/

PR #499, opened 2024-02-28
Add Chart::Events testsuite

schaumb:


src/base/util/eventdispatcher.h
```diff
@@ -69,6 +69,13 @@ class EventDispatcher
 
 	bool destroyEvent(const event_ptr &event);
 
+	void registerOnEachEvent(const std::uint64_t &id,
```
simzer, line 72: Would be better to handle this at gui events

simzer (approved):

simzer (approved):

PR #509, opened 2024-03-13
Chart onChange and rendering connection moved to TS API

simzer:


schaumb:
When a resize happens, it needs to call Chart::doChange on the "ts" side. Would you like to change that in this PR?

simzer:
Checking size change inside wasm was obsolate, since size in wasm only got updated when needsUpdate was already queried.

simzer:
@schaumb No, on a second tought, i don't wan't to change the current size handling, because currently we force a redraw immediately if the canvas size changed. If we set the needsUpdate flag instead, it will only be redrawn when the next rendering scheduled, hance there migth be a visible delay on rendering refresh when we rescale the canvas.

schaumb (approved):

PR #511, opened 2024-03-14
Wrong y axis label position for line/circle in case of both axis dimension

schaumb:
The best solution would be to put the markers in the middle instead of at the top of the positions where the bars would be

### DIFF
--- a/src/chart/generator/marker.cpp
+++ b/src/chart/generator/marker.cpp
@@ -64,6 +64,9 @@ Marker::Marker(const Options &options,
 	}
 
 	auto horizontal = options.isHorizontal();
+	bool lineOrCircle = options.geometry == ShapeType::line
+	                 || options.geometry == ShapeType::circle;
+
 	position.x = size.x = getValueForChannel(channels,
 	    ChannelId::x,
 	    data,
@@ -71,7 +74,8 @@ Marker::Marker(const Options &options,
 	    options.subAxisOf(ChannelId::x),
 	    !horizontal && stackInhibitingShape);
 
-	spacing.x = (horizontal && options.getChannels().anyAxisSet()
+	spacing.x = ((horizontal || lineOrCircle)
+	                && options.getChannels().anyAxisSet()
 	                && channels.at(ChannelId::x).isDimension())
 	              ? 1
 	              : 0;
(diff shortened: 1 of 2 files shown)


schaumb:
This is not working yet, because orientation change is not recognized as an animation step (jumps)

src/chart/generator/marker.cpp
```diff
@@ -64,14 +64,18 @@ Marker::Marker(const Options &options,
...
+
 	position.x = size.x = getValueForChannel(channels,
 	    ChannelId::x,
 	    data,
 	    stats,
 	    options.subAxisOf(ChannelId::x),
 	    !horizontal && stackInhibitingShape);
 
-	spacing.x = (horizontal && options.getChannels().anyAxisSet()
+	spacing.x = ((horizontal || lineOrCircle)
```
simzer, line 77: This fixes the line/circle case, but this would fix the rectangle case as well:
```cpp
	spacing.x = (options.getChannels().anyAxisSet()
	                && channels.at(ChannelId::x).isDimension())
	              ? 1
	              : 0;
```

simzer (changes requested):

src/chart/generator/marker.cpp
simzer: I think it was a problem to even consider orientation here. Only the axis type (dimension or not) counts

schaumb: But this will fail on some test cases where it is intended to work like before. 
kép
kép
kép

simzer: Yep, they should look like as they look on the right side. ->

simzer: Oh, I see. The result is better, but the transition is problematic!

simzer (approved):

simzer (approved):

src/chart/generator/plot.cpp
```diff
@@ -159,8 +159,15 @@ void Plot::generateMarkers(const Data::DataTable &table)
 	}
 	clearEmptyBuckets(mainBuckets, true);
 	clearEmptyBuckets(subBuckets, false);
-	linkMarkers(mainBuckets, true);
+	auto conn = linkMarkers(mainBuckets, true);
```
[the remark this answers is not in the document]
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/514/commits/36dcc05721a691da3f905bb723cd77c7df061f5c

PR #512, opened 2024-03-18
Rendering fixed, update separated to update and render on wasm api

simzer:


src/apps/weblib/interface.cpp
```diff
@@ -373,15 +369,24 @@ void Interface::update(ObjectRegistry::Handle chart,
...
-		widget->onDraw(ptr);
-		ptr->frameEnd();
-	}
+void Interface::render(ObjectRegistry::Handle chart,
+    ObjectRegistry::Handle canvas,
+    double width,
+    double height)
+{
+	auto &&widget = objects.get<UI::ChartWidget>(chart);
+	auto ptr = objects.get<Vizzu::Main::JScriptCanvas>(canvas);
```
schaumb, line 380: make it `auto &&` to force not copy the `shared_ptr`

src/apps/weblib/interface.cpp
```diff
@@ -373,15 +369,24 @@ void Interface::update(ObjectRegistry::Handle chart,
...
+
+	ptr->frameBegin();
+
+	const Geom::Size size{width, height};
+	widget->onUpdateSize(size);
```
schaumb, line 385: inline `size` variable, or make the c++ Interface parameter as `Geom::Size` replacement of 2 doubles (width, height).

src/apps/weblib/ts-api/chart.ts
```diff
@@ -75,35 +81,55 @@ export class Chart implements ChartInterface {
...
 		}
 		this._plugins.hook(Hooks.render, ctx).default((ctx) => {
-			if (ctx.size.x >= 1 && ctx.size.y >= 1 && ctx.timeInMSecs !== null && ctx.renderer) {
-				const render = ctx.force || (ctx.enable && this._needsUpdate)
-				ctx.renderer.canvas = this._cCanvas
-				this._module.registerRenderer(this._cCanvas, ctx.renderer)
-				this._cChart.update(this._cCanvas, ctx.size.x, ctx.size.y, ctx.timeInMSecs, render)
-				this._module.unregisterRenderer(this._cCanvas)
-				this._needsUpdate = false
+			if (ctx.size.x >= 1 && ctx.size.y >= 1 && ctx.renderer) {
```
schaumb, line 116: negate it and make it an early return.

src/apps/weblib/ts-api/chart.ts
```diff
@@ -75,35 +81,55 @@ export class Chart implements ChartInterface {
...
+				const shouldRender =
+					ctx.control === RenderControlMode.forced ||
+					(ctx.control === RenderControlMode.allowed && ctx.changed)
+				if (shouldRender) {
```
schaumb, line 120: negate it and make it an early return.

src/apps/weblib/ts-api/plugins.ts
```diff
@@ -10,6 +10,8 @@ import { Canvas } from './module/canvas.js'
 export enum Hooks {
 	/** Called once on startup for start the rendering loop. */
 	start = 'start',
+	/** Called when updateing the chart due to time change. */
```
schaumb, line 13: typo in `updateing`

src/apps/weblib/interface.cpp
```diff
@@ -356,11 +356,7 @@ void Interface::setLogging(bool enable)
 }
 
 void Interface::update(ObjectRegistry::Handle chart,
-    ObjectRegistry::Handle canvas,
-    double width,
-    double height,
-    double timeInMSecs,
-    bool render)
+    double timeInMSecs)
```
schaumb, line 359: Make the C++ interface parameter to C++ class (`std::chrono::duration<double, std::milli>`)

simzer: Currently we have the C->C++ type conversions in interface.cpp. The idea was that this is a C interface but the implementation is encapsulated in a c++ class, and the standalone functions are just thin wrappers. If we want to change the concept, we should move all type conversations to cinterface.cpp

src/apps/weblib/ts-api/chart.ts
simzer: what do you mean?

schaumb: current version:
```
if (...) {
  // do stuff
}
// nothing
```
expected:
```
if (!(...))
  return; // nothing
  
// do stuff
```

This will cause not-nested tabulators

src/apps/weblib/interface.cpp
schaumb: I previously changed the `setChartFilter`, `getRecordValue`, and `set/get AnimControlValue` (string_view). I think this concept can be reached step-by-step with small modifications.

simzer: Let's do this in one step, it's more efficient: 
interface.cpp method arguments should be c++ types
[#513](https://github.com/vizzuhq/vizzu-lib/issues/513)

dovicsin (approved):

PR #515, opened 2024-03-22
Refactor based modifications - MultiDim, DataCube, DataStat

schaumb:


src/data/table/datatable.cpp
```diff
@@ -81,20 +80,16 @@ DataTableOld::DataIndex DataTableOld::addTypedColumn(
 
 		addRow(row);
 	}
-
-	return getIndex(ColumnIndex(colIndex));
 }
 
-DataTableOld::DataIndex DataTableOld::addColumn(
-    const std::string &name,
+void DataTableOld::addColumn(const std::string &name,
```
simzer, line 85: returns from a void function. same in addColumn

schaumb: that's OK, it's well-defined.

simzer: The issue is not how defined it is, but the readability of the code. I think it's not a good practice to return with void.

simzer: But since this source file will be removed soon, it's not necessary to change it.

simzer (approved):

PR #521, opened 2024-04-29
Reserve marker's space

schaumb:


src/chart/speclayout/bubblechartbuilder.h
```diff
@@ -46,26 +41,23 @@ void BubbleChartBuilder::setupVector(std::vector<Item> &items,
...
-		for (const auto &item : level.second)
-			sizes.push_back(
-			    std::max(0.0, items[item.second].sizeFactor));
+		sizes.reserve(std::size(level));
+		for (const auto &item : level)
+			sizes.push_back(std::max(0.0, item->sizeFactor));
 
 		const BubbleChart subChart(sizes, c.boundary());
 
 		size_t subCnt = 0;
```
schaumb, line 50: Move to the for's init

src/chart/generator/plot.cpp
```diff
@@ -112,19 +112,33 @@ bool Plot::isEmpty() const
 
 void Plot::generateMarkers()
 {
-	for (auto &&index : getDataCube()) {
-		auto markerIndex = markers.size();
+	if (!getDataCube().empty()) {
+		auto &&[k, v] = getDataCube().combinedSizeOf(
+		    options->mainAxis().dimensionIds);
+		mainBuckets.resize(k, std::vector<Marker *>(v));
```
schaumb, line 118: Check that when the buckets clearable

src/chart/generator/plotptr.h
```diff
@@ -2,12 +2,74 @@
...
 namespace Vizzu::Gen
 {
 class Plot;
-
 using PlotPtr = std::shared_ptr<Plot>;
+
+class Marker;
+struct Buckets
+{
+	std::vector<Marker *> markers;
```
schaumb, line 16: Check that can it be replace to unique ptr arr

src/chart/rendering/colorbuilder.h
```diff
@@ -26,7 +26,7 @@ struct ColorBuilder
 	    const Gen::ColorBase &colorBase) const;
 
 private:
-	LighnessRange lighnessRange{0.4, -0.4};
+	LighnessRange lighnessRange = LighnessRange::Raw(-0.4, 0.4);
```
schaumb, line 29: @simzer Is this the expected working?

src/base/text/numberscale.cpp
```diff
@@ -36,7 +37,7 @@ NumberScale::NumberScale(std::string s)
 	else {
```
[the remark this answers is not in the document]
schaumb: it seems that already a "move"

src/chart/rendering/colorbuilder.h
schaumb: If I swap the arguments, all test cases run without any difference.

schaumb: So this parameter (this construction) is unused. The only constructor initializes this parameter through argument, not by default.

schaumb: kép
The default argument is (-0.4, 0.4).

src/base/type/booliter.h
```diff
@@ -0,0 +1,23 @@
...
+#define TYPE_BOOLITER_H
+
+#include <array>
+
+namespace Type
+{
+struct Bools
+{
+	constexpr static std::array boolArr = {false, true};
+	bool max;
```
schaumb, line 11: Default Val: true

src/chart/animator/keyframe.cpp
```diff
@@ -71,51 +71,34 @@ void Keyframe::prepareActual()
...
-	}
-	for (auto &item : origTMI) {
-		auto iter = smi.find(item.first);
-		if (iter != smi.end())
-			smi.insert(std::make_pair(item.first, item.second));
+	if (!smi.empty()) copyTarget();
+
+	for (auto &tmi = target->getMarkersInfo(); auto &&item : smi)
+		if (origTMI.contains(item.first))
+			tmi.insert({item.first, item.second});
```
schaumb, line 80: This insert has no reason if copyTarget copying the markerinfo

src/chart/generator/axis.cpp
```diff
@@ -20,10 +20,12 @@ Geom::Point MeasureAxises::origo() const
 
 MeasureAxis::MeasureAxis(Math::Range<double> interval,
     std::string &&unit,
+    const std::string_view &measName,
     std::optional<double> step) :
     enabled(true),
     range(interval),
     unit(std::move(unit)),
+    origMeasureName(measName),
```
schaumb, line 28: What is happens if datatable has 4 measure, we have a plot at least one measure (which use SSO), then we add another measure?

src/chart/generator/marker.cpp
```diff
@@ -40,20 +44,24 @@ Marker::Marker(const Options &options,
...
+	    data.getId(options.mainAxis().dimensionsWithLevel(), index);
 
+	Data::MarkerId *subAxisId{};
 	if (options.geometry == ShapeType::area) {
-		Data::SeriesList subIds(options.subAxis().dimensionIds);
-		Data::SeriesList &&stackIds =
-		    subIds.split_by(options.mainAxis().dimensionIds);
-		subId = data.getId(subIds, index);
-		stackId = data.getId(stackIds, index);
+		auto &&subAxis = options.subAxis();
```
schaumb, line 55: Move to if init

src/chart/generator/marker.cpp
```diff
@@ -135,31 +138,25 @@ void Marker::resetSize(bool horizontal)
...
-	return res;
+		nextSubMarkerIdx->value += offset;
 }
 
 Conv::JSONObj &&Marker::appendToJSON(Conv::JSONObj &&jsonObj) const
 {
-	return std::move(jsonObj)("categories",
-	    cellInfo.categories)("values", cellInfo.values)("index", idx);
+	if (cellInfo) return std::move(jsonObj).merge(cellInfo->json);
+	return std::move(jsonObj);
```
schaumb, line 151: Check which cases when we run here

src/dataframe/old/datatable.cpp
```diff
@@ -295,34 +313,67 @@ MarkerId DataCube::getId(const SeriesList &sl,
...
-
+	std::vector<std::string_view> res(sl.size());
 	for (std::size_t ix{}; ix < dim_reindex.size(); ++ix) {
-		auto &&name = dim_reindex[ix].first;
-		auto cats = df->get_categories(name);
-		auto cix = index.old[ix];
-		my_res.categories.emplace_back(name,
-		    cix < cats.size() ? cats[cix] : std::string_view{});
+		auto &&[name, cats, size] = dim_reindex[ix];
+		if (auto &&i = sl.find(name); i != SeriesList::npos) {
```
schaumb, line 323: Is this N*log (N) necessary?

schaumb: Not. Refactored to linear.

src/chart/generator/marker.cpp
schaumb: @simzer This is a behavior change. Some markers are disabled by both plots (source/target) but should be drawn because of a connection (previous or next enabled marker) on a line/area plot. For these markers, the cellInfo is not generated. This means disabled markers' info is not sent on a 'plot-marker-draw' event.

schaumb: (note: if the marker "needs" a tooltip, the cellInfo is generated)

src/base/anim/interpolated.h
```diff
@@ -82,145 +83,129 @@ template <typename Type> class Interpolated
...
 		    && values[0].value == other;
 	}
 
 	bool operator<(const Interpolated<Type> &other) const
 	{
-		if (count != 1 && other.count != 1)
+		if (has_second && other.has_second)
 			throw std::logic_error("cannot compare weigthed pairs");
-		if (count >= other.count) return false;
+		if (!other.has_second) return false;
```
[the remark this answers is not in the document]
schaumb: This is the same condition as before.

| this ↓ .has_second → other  | false | true |
| ------------------------------------- | --------- | --- |
| false | Handled by this if | other.count was bigger than count (false >= true) |
| true | Handled by this if | Previous if condition, not applicable |

src/chart/speclayout/bubblechart.cpp
```diff
@@ -11,10 +11,8 @@ BubbleChart::BubbleChart(const std::vector<double> &sizes,
 {
 	markers.reserve(sizes.size());
 
-	for (auto j = 0U; j < sizes.size(); ++j) {
-		const double radius = std::sqrt(sizes[j]);
-		markers.emplace_back(j, radius);
-	}
+	for (auto j = 0U; j < sizes.size(); ++j)
+		markers.emplace_back(j, std::sqrt(std::max(0.0, sizes[j])));
```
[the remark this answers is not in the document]
schaumb: The name (radius) did not say why it should be used the sqrt. That's a mathematical computation without any knowledge of information.

src/base/anim/interpolated.h
```diff
@@ -82,145 +83,129 @@ template <typename Type> class Interpolated
...
 	[[nodiscard]] const Type &get() const
 	{
-		if (count == 1) return values[0].value;
+		if (!has_second) return values[0].value;
 
 		throw std::logic_error("Invalid Weigthed Pair");
 	}
 
-	[[nodiscard]] const Weighted<Type> &get(uint64_t index) const
+	[[nodiscard]] const Weighted<Type> &get(bool index) const
```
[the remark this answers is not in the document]
schaumb: I agree. Is this enum OK?
```cpp
enum Index : bool {
  first = false,
  secondIfExists = true
};
```

src/apps/weblib/objectregistry.h
```diff
@@ -1,23 +1,35 @@
...
 
 namespace Vizzu
 {
+namespace UI
+{
+class ChartWidget;
+}
+
+struct Snapshot;
+struct Animation;
```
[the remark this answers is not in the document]
schaumb: OK

src/chart/generator/marker.cpp
simzer: I think it's ok, this whole part will be changed with new marker morph.

src/chart/rendering/markerrenderer.h
```diff
@@ -33,13 +33,17 @@ class MarkerRenderer : public DrawingContext
 	    bool isLine) const;
 	void drawLabel(Gfx::ICanvas &canvas,
 	    const AbstractMarker &abstractMarker,
-	    size_t index) const;
```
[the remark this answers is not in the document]
schaumb: The "index" variable name is OK, but the type is not OK.

src/apps/weblib/objectregistry.h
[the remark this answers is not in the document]
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/521/commits/99cfeec0360525034552c5655261b289b5b04bc9

src/chart/speclayout/bubblechart.cpp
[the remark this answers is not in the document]
simzer: radius together with sqrt gives the idea that size value used as the area of the circle marker, not as the radius, which would be a mistake from data visualization point of view.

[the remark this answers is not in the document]
schaumb: What if I rename the function argument from `sizes` to `circleAreas`, and specMarker constructor argument from `size` to `radiusOrAreaFactor`? Is it enough self documentation for the transition?

src/base/anim/interpolated.h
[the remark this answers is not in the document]
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/521/commits/0ddd834ff61eb0bca480c9ba479f64b54f91ac49

src/chart/animator/keyframe.cpp
```diff
@@ -64,58 +68,47 @@ void Keyframe::prepareActual()
...
 {
-	auto &origTMI = target->getMarkersInfo();
+	const auto &origTMI = target->getMarkersInfo();
 	auto &smi = source->getMarkersInfo();
-	for (auto &item : smi) {
-		auto iter = origTMI.find(item.first);
-		if (iter != origTMI.end()) {
-			copyTarget();
-			target->getMarkersInfo().insert(
-			    std::make_pair(item.first, item.second));
```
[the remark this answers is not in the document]
schaumb: the markersInfo map is unique by the key. So if the key is in the map, it must contain the same markerId.

src/base/anim/interpolated.h
```diff
@@ -41,6 +41,20 @@ template <typename Type> class Weighted
 	[[nodiscard]] bool hasValue() const { return weight > 0.0; }
 };
 
+enum class InterpolateIndex : bool {
+	first = false,
+	secondIfExists = true
```
simzer, line 46: Reviewing the usage of this enum, it seams to me, that the correct naming would be `first` and `second`, because `secondIfExists` can refer to the first value only in the get() method. So I think the get() method should have an adequate name instead, like `get_or_first()` or something like that.

schaumb: https://github.com/vizzuhq/vizzu-lib/pull/521/commits/3402a37a53b3b25302b9f72200da369ede94aeae

simzer (approved):

PR #522, opened 2024-05-13
Set font size when resize happens

### LINKED ISSUE #315: No Layout/font recalc on resize, only when animate called
Default style parameters recalculated only when animate called, causing strange effect after canvas resize, that the chart fonts, paddings adjust only when the next animation called.

Peter:
Chart elements don't scale proportionally when the full-screen mode is turned on (also won't shrink when it's turned off) 
https://github.com/vizzuhq/ipyvizzu-story/assets/10341970/fcc5dfae-d577-4f53-837a-53a24f7f45a3

schaumb:
Resolves #315


simzer (changes requested):
Please write note to changelog

simzer (approved):

PR #523, opened 2024-05-14
Add dataframe exceptions

schaumb:
This modification results: data must always come before configuration.


schaumb:
After the https://github.com/vizzuhq/vizzu-lib/pull/521 merge, the CI will be green.

PR #526, opened 2024-05-16
Dim--dim on x y

schaumb:


src/chart/generator/marker.cpp
```diff
@@ -75,7 +76,12 @@ Marker::Marker(const Options &options,
 	    index,
 	    horizontal ? &mainId->value : subAxisId);
 
-	spacing.x = (horizontal || (lineOrCircle && !polar))
+	spacing.x = (horizontal || (lineOrCircle && !polar)
+	                || (channels.at(ChannelId::y).isDimension()
+	                    && channels.at(ChannelId::y).hasDimension()
```
simzer, line 81: Why do we need hasDimension(), shouldnt isDimension be enought?

src/chart/generator/marker.cpp
```diff
@@ -88,7 +94,12 @@ Marker::Marker(const Options &options,
 	    index,
 	    !horizontal ? &mainId->value : subAxisId);
 
-	spacing.y = (!horizontal || lineOrCircle)
+	spacing.y = (!horizontal || lineOrCircle
+	                || (channels.at(ChannelId::x).isDimension()
+	                    && channels.at(ChannelId::x).hasDimension()
```
simzer, line 99: This expression is getting too complex. Maybe extract newly added portion to a local variable?

src/chart/generator/plot.cpp
```diff
@@ -129,6 +129,11 @@ Buckets Plot::generateMarkers()
 	for (auto &&[ix, mid] : options->markersInfo)
 		map.emplace(mid, ix);
 
+	bool rectangleSpacing{true};
+	if (auto &&sp = style.plot.marker.rectangleSpacing;
+	    sp && sp->get().has_value())
+		rectangleSpacing = *sp->get() > 0.0;
+
```
simzer, line 136: Why is this needed? Cannot we handle this only at rendering? Marker::spacing supposed to be a style independent factor

src/chart/options/options.cpp
```diff
@@ -115,6 +115,9 @@ void Options::drilldownTo(const Options &other)
 	for (auto &&dim : other.getChannels().getDimensions())
 		if (!getChannels().isSeriesUsed(dim))
 			stackChannel.addSeries(dim);
+	if (stackChannel.isDimension()
+	    && geometry == ShapeType::rectangle)
+		this->align = Base::Align::Type::stretch;
```
simzer, line 120: It smells a bit "workaroundy". Stretch does the work, but it's purpose is complety different. 
I feel like that instead, we would need an `bool spacing` member in `Channel` which gives controls to manually swich on/off the spacing.

src/chart/generator/marker.cpp
schaumb: isDimension() && !hasDimension means == empty(). When it is an empty channel, there should be no spacing

src/chart/generator/plot.cpp
simzer: After checking the test cases, we concluded that this is not needed.

src/chart/options/options.cpp
schaumb: I noted. I'll not rewrite it in this PR but handle it at the axis refactor.

simzer (approved):

PR #530, opened 2024-05-22
Fix empty datacube calculation

schaumb:


simzer (dismissed):

simzer:
Would it make sense to add a test case covering this fix?

simzer (approved):

simzer (approved):

PR #540, opened 2024-06-12
Fix treemap -> tablemap

schaumb:


src/chart/speclayout/tablechart.h
```diff
@@ -18,14 +18,25 @@ template <typename Item>
...
-	    singleColumn ? 1 : static_cast<ssize_t>(ceil(sqrt(size)));
+	if (size == 0) return;
+
+	ssize_t rowsize{1};
+	if (!singleColumn) {
+		rowsize = static_cast<ssize_t>(ceil(sqrt(size)));
+		auto sq = static_cast<ssize_t>(ceil(sqrt(
+		    16.0 * static_cast<double>(rowsize * rowsize) / 9.0)));
+
+		auto rem = rowsize - 1 - (size - 1) % rowsize;
```
simzer, line 33: Good idea, but it will cause strange rearrangements in case the user filters out elements from the chart. 
If I understand it correctly, I have a 7x11 element in 7 rows, if I filter one out, it will rearrange it to 9 rows.

schaumb: The same happens in the origin version when it filters out enough markers to decrease the active markers count square root, or the filtered markers count does not match the sqrt of valid marker size. The rearrangement is always strange when dimension sizes are different than the sqrt (just think about 4 * 11).

This code at least tries to minimize the "empty" spaces.

simzer: The change is bigger with the new version, on the other hand it was not a probllem imho, that it did not fill the fill the space, we are just listing the markers here, it's like word wrap for a text paragraph, the last line won't go till the end of line.

Let's ask for a second opinion. @andraskangyal, please read the discussion, then check out the difference:
Old: https://jsfiddle.net/simzer/um6jh5cv/6/
new: https://jsfiddle.net/simzer/um6jh5cv/7/

andraskangyal: If I'm looking at the 'Old' one, I feel it's better because the line shifting animation is clearer. But I don't know is there any other consequences too behind the new concept.(?)

simzer: New:
pro: we have full last line more times, and we don't have only few or one markers in the last line.
contra: the row count goes up-down-up-down while we only adding the markers.

simzer: Andris's decision: we go with the new one. but 16:9 should be changed to 1:1

simzer, line 33: it would be more readable if this would be moved to a separate function with a descriptive name.

src/chart/speclayout/tablechart.h
```diff
@@ -18,14 +18,25 @@ template <typename Item>
...
 
-	auto rowsize =
-	    singleColumn ? 1 : static_cast<ssize_t>(ceil(sqrt(size)));
+	if (size == 0) return;
+
+	ssize_t rowsize{1};
+	if (!singleColumn) {
+		rowsize = static_cast<ssize_t>(ceil(sqrt(size)));
+		auto sq = static_cast<ssize_t>(ceil(sqrt(
+		    16.0 * static_cast<double>(rowsize * rowsize) / 9.0)));
```
simzer, line 31: we should use 1:1 ratio to 
- prevent markers to be too narrow on portrait layout charts
- fatter markers look better (according to @andraskangyal)

simzer (changes requested):

src/chart/speclayout/tablechart.h
schaumb: 1:1 has doesn't make sense. Sqrt makes 1:1

schaumb: Or this means keepaspectratio should be true?

simzer: It means that now the grid will be between 1:1 -- 16:9. We should change that to a symmetric range around 1:1 to achieve the 2 points I wrote in the first comment. E.g. 3:4 -- 4:3

simzer (changes requested):

src/chart/speclayout/tablechart.h
schaumb: How can we decide that it should be portrait or landscape? There are no information about canvas aspect ratio here.

schaumb: If we want to make it 4:3 then 1:1- 4:3 is the only option here (3:4-4:3 doesn't make any sense, because no x-y order exists). The other option is the keepAspectRatio

simzer: OK, so the 2 problems I mentioned in the comment are also a disadvantage of this solution which we did not consider during the decision. @andraskangyal please give this another thought.

schaumb: note: the current solution on portrait aspect ratio, the markers are portrait too (so taller than wider), so currently those are not "fatter" at all.

simzer: Fatter then the proposed solution.

schaumb: but with 16:9 -> 1:1 and keepAspectRatio can solve this.

simzer: But it introduces yet another problem, we won't fill up the available plot area space

simzer: We discussed this again with @andraskangyal, considering the disadvantages of all solutions: 
- old version: less marker in the last row
- new version: makes the markers narrower, which is not desired in portrait layout
- new version + keep aspect ratio: makes empty space above/below or left/right side of the chart
we should keep the old version.

simzer: Also asked @dovicsin to research possible visualizations to represent pure dimensional data. Maybe table chart not our best option for this job on the first place.

simzer (approved):

PR #550, opened 2024-07-08
HighResolution as parameter

schaumb:


schaumb:
I'll reopen when  markerconnection is merged

PR #553, opened 2024-07-18
Fix sort pointers and doubles

schaumb:


src/base/anim/interpolated.h
```diff
@@ -33,9 +34,13 @@ template <typename Type> class Weighted
 		return value == other.value && weight == other.weight;
 	}
 
+	template <typename Comparator =
```
simzer, line 37: `Comparator` could be named `Less`

src/base/math/floating.h
```diff
@@ -1,20 +1,65 @@
...
+        std::bit_cast<std::uint64_t>(F{0.0}) == std::uint64_t{}
+            && std::bit_cast<std::uint64_t>(F{-0.0})
+                       + std::bit_cast<std::uint64_t>(F{-0.0})
+                   == std::uint64_t{}>{};
+
+constexpr auto inline is_zero = [](auto value)
+{
+	using F = decltype(value);
+	static_assert(std::floating_point<F>);
+	if constexpr (auto v = can_be_used_as_short_check<F>; v()) {
```
simzer, line 52: Please put a comment with link to a description.

simzer (approved):

src/base/math/floating.h
schaumb: I'll revert this "hack" because on wasm the float64 and int64 have different "assembly" sets, and on wasm target, this will be not faster.

Original link: https://www.agner.org/optimize/optimizing_assembly.pdf (147-148 page), and https://stackoverflow.com/a/46772747

schaumb: https://github.com/vizzuhq/vizzu-lib/pull/555/commits/fafd45e1db663964cf1ac0b117b926414f0524d7

src/base/anim/interpolated.h
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/555/commits/fafd45e1db663964cf1ac0b117b926414f0524d7

PR #554, opened 2024-07-18
Canvas line resolution

schaumb:


src/apps/weblib/cinterface.h
```diff
@@ -68,6 +68,10 @@ extern void vizzu_render(APIHandles::Chart chart,
     APIHandles::Canvas canvas,
     double width,
     double height);
+extern void vizzu_setLineResolution(APIHandles::Canvas canvas,
+    double dMax,
```
simzer, line 72: we could use readable names.
Distance max, CurveHeight max

src/apps/weblib/ts-api/cvizzu.types.d.ts
```diff
@@ -91,7 +91,8 @@ export interface CVizzu {
 	_vizzu_pointerLeave(chart: CChartPtr, canvas: CCanvasPtr, pointerId: number): void
 	_vizzu_wheel(chart: CChartPtr, canvas: CCanvasPtr, delta: number): void
 	_vizzu_setLogging(enable: boolean): void
-	_vizzu_update(chart: CChartPtr, time: number): void
+	_vizzu_update(chart: CChartPtr, time: number): CString
```
simzer, line 94: void return

simzer (approved):

src/apps/weblib/ts-api/cvizzu.types.d.ts
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/555/commits/8a05e8f4845349854efa9b8d8f8cd3e1524c6b06

PR #556, opened 2024-07-29
Set version to 0.12.0

veghdev:


tools/ci/run/init-py.sh
```diff
@@ -11,5 +11,6 @@ test -f ~/.netrc && chmod u+rw,u-x,go-rwx ~/.netrc
 
 python3.10 -m venv --copies ".venv" || python3 -m venv --copies ".venv"
 source .venv/bin/activate
+pip3 install "setuptools<72.0.0"
```
schaumb, line 14: the next line is pip, this is pip3. It should be the same (all pip3 or all pip)

schaumb (approved):

schaumb (approved):

PR #563, opened 2024-07-30
Change marker id to string

schaumb:


schaumb:
I'll create another PR

PR #564, opened 2024-07-31
Fix clang-tidy(-18) issues

schaumb:
- parent class function hides same member function
- Replace CHAR_BITS and M_PI to C++ replacements
- Remove void function returns
- move rvalue references/perfect forward universal references
- move type_traits to C++17 '_v' suffixes
- include all includes to C++
- Remove nested '?:' ternary operators
- Remove implicit inlines
- Make all enum base to uint8_t
- Remove some unnecessary namespace prefix
- Change enable_if to requires
- std::ignore some argument and return values
- header guard to canvas.h


src/chart/animator/planner.cpp
```diff
@@ -106,17 +122,21 @@ void Planner::createPlan(const Gen::Plot &source,
 		addMorph(SectionId::color, step);
 		addMorph(SectionId::coordSystem, std::max(step, posDuration));
 
-		const auto &geomEasing =
-		    srcOpt->geometry == Gen::ShapeType::circle   ? in3
-		    : trgOpt->geometry == Gen::ShapeType::circle ? out3
-		    : srcOpt->geometry == Gen::ShapeType::line   ? in3
-		    : trgOpt->geometry == Gen::ShapeType::line   ? out3
-		                                                 : inOut5;
+		const ::Anim::Easing *geomEasing{&inOut5};
```
simzer, line 125: if else

simzer (approved):

simzer (approved):

src/chart/animator/planner.cpp
schaumb: The calng-tidy says that there are identical if-branches.

PR #566, opened 2024-08-21
Fix dimension columns aggregation with non-existent values

schaumb:


schaumb:
I'll revert that I removed the 'exists' because it has a valid meaning

simzer (approved):

PR #568, opened 2024-08-26
Add translateY style parameter to legend + test

schaumb:


schaumb:
I can add the following patch to this PR if needed: It adds 'yOverflow' to the Legend draws event.

```
Subject: [PATCH] yOverflow in Legend event
---
Index: src/chart/rendering/drawlegend.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/chart/rendering/drawlegend.cpp b/src/chart/rendering/drawlegend.cpp
--- a/src/chart/rendering/drawlegend.cpp	(revision 3d3fc0a53c69be54e1abb0e9dd71e5ea001a209e)
+++ b/src/chart/rendering/drawlegend.cpp	(date 1724678063245)
@@ -39,6 +39,7 @@
 	    .canvas = canvas,
 	    .titleRect = titleRect,
 	    .markerWindowRect = markerWindowRect,
+	    .yOverflow = {},
 	    .yOffset = {},
 	    .type = channelType,
 	    .weight = weight,
@@ -49,16 +50,17 @@
 	    .dimension = plot->axises.at(channelType).dimension,
 	};
 
-	auto yOverflow =
+	info.yOverflow =
 	    markersLegendFullSize(info) - markerWindowRect.height();
-	if (std::signbit(yOverflow)) yOverflow = 0.0;
-	info.yOffset = style.translateY->get(yOverflow, info.itemHeight);
+	if (std::signbit(info.yOverflow)) info.yOverflow = 0.0;
+	info.yOffset =
+	    style.translateY->get(info.yOverflow, info.itemHeight);
 
 	DrawBackground{{ctx()}}.draw(canvas,
 	    legendLayout,
 	    style,
 	    *events.background,
-	    Events::Targets::legend(channelType));
+	    Events::Targets::legend(channelType, info.yOverflow));
 
 	canvas.save();
 	canvas.setClipRect(contentRect);
@@ -91,7 +93,9 @@
 		        title.value,
 		        style.title,
 		        *events.title,
-		        Events::Targets::legendTitle(title.value, info.type),
+		        Events::Targets::legendTitle(title.value,
+		            info.type,
+		            info.yOverflow),
 		        {.alpha = title.weight * info.weight * mul});
 	    });
 }
@@ -133,7 +137,8 @@
 			            info.dimension.category,
 			            value.second.categoryValue,
 			            value.second.categoryValue,
-			            info.type),
+			            info.type,
+			            info.yOverflow),
 			        {.alpha = double{
 			             alpha && Math::FuzzyBool{weighted.weight}}});
 		    });
@@ -186,7 +191,8 @@
 	auto markerElement =
 	    Events::Targets::legendMarker(info.dimension.category,
 	        categoryValue,
-	        info.type);
+	        info.type,
+	        info.yOverflow);
 
 	if (events.marker->invoke(
 	        Events::OnRectDrawEvent(*markerElement, {rect, false}))) {
@@ -245,7 +251,9 @@
 	    text,
 	    style.label,
 	    *events.label,
-	    Events::Targets::measLegendLabel(text, info.type),
+	    Events::Targets::measLegendLabel(text,
+	        info.type,
+	        info.yOverflow),
 	    {.alpha = info.measureWeight * plusWeight});
 }
 
@@ -283,7 +291,8 @@
 	info.canvas.setLineWidth(0);
 
 	auto barElement =
-	    Events::Targets::legendBar(Gen::ChannelId::color);
+	    Events::Targets::legendBar(Gen::ChannelId::color,
+	        info.yOverflow);
 
 	if (events.bar->invoke(
 	        Events::OnRectDrawEvent(*barElement, {rect, false}))) {
@@ -315,7 +324,8 @@
 	info.canvas.setLineWidth(0);
 
 	auto barElement =
-	    Events::Targets::legendBar(Gen::ChannelId::lightness);
+	    Events::Targets::legendBar(Gen::ChannelId::lightness,
+	        info.yOverflow);
 
 	if (events.bar->invoke(
 	        Events::OnRectDrawEvent(*barElement, {rect, false}))) {
@@ -336,8 +346,8 @@
 	    Gfx::Color::Gray(0.8) * info.measureWeight);
 	info.canvas.setLineWidth(0);
 
-	auto barElement =
-	    Events::Targets::legendBar(Gen::ChannelId::size);
+	auto barElement = Events::Targets::legendBar(Gen::ChannelId::size,
+	    info.yOverflow);
 
 	if (events.bar->invoke(
 	        Events::OnRectDrawEvent(*barElement, {rect, false}))) {
Index: src/chart/main/events.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/chart/main/events.h b/src/chart/main/events.h
--- a/src/chart/main/events.h	(revision 3d3fc0a53c69be54e1abb0e9dd71e5ea001a209e)
+++ b/src/chart/main/events.h	(date 1724678063222)
@@ -208,16 +208,19 @@
 		struct Legend : Element
 		{
 			Gen::ChannelId channel;
+			double yOverflow;
 
-			explicit Legend(Gen::ChannelId channel) :
+			explicit Legend(Gen::ChannelId channel,
+			    double yOverflow) :
 			    Element("legend"),
-			    channel(channel)
+			    channel(channel),
+			    yOverflow(yOverflow)
 			{}
 
 			void appendToJSON(Conv::JSONObj &&jsonObj) const override
 			{
-				Element::appendToJSON(
-				    std::move(jsonObj)("channel", channel));
+				Element::appendToJSON(std::move(jsonObj)("channel",
+				    channel)("yOverflow", yOverflow));
 			}
 		};
 
@@ -301,9 +304,9 @@
 			return std::make_unique<Axis>(horizontal);
 		}
 
-		static auto legend(Gen::ChannelId channel)
+		static auto legend(Gen::ChannelId channel, double yOverflow)
 		{
-			return std::make_unique<Legend>(channel);
+			return std::make_unique<Legend>(channel, yOverflow);
 		}
 
 		static auto marker(const Gen::Marker &marker)
@@ -366,46 +369,57 @@
 		    const std::string_view &categoryName,
 		    const std::string_view &categoryValue,
 		    const std::string &label,
-		    Gen::ChannelId channel)
+		    Gen::ChannelId channel,
+		    double yOverflow)
 		{
 			return std::make_unique<CategoryInfo<Text<LegendChild>>>(
 			    categoryName,
 			    categoryValue,
 			    label,
 			    "label",
-			    channel);
+			    channel,
+			    yOverflow);
 		}
 
 		static auto measLegendLabel(const std::string &label,
-		    Gen::ChannelId channel)
+		    Gen::ChannelId channel,
+		    double yOverflow)
 		{
 			return std::make_unique<Text<LegendChild>>(label,
 			    "label",
-			    channel);
+			    channel,
+			    yOverflow);
 		}
 
 		static auto legendTitle(const std::string &title,
-		    Gen::ChannelId channel)
+		    Gen::ChannelId channel,
+		    double yOverflow)
 		{
 			return std::make_unique<Text<LegendChild>>(title,
 			    "title",
-			    channel);
+			    channel,
+			    yOverflow);
 		}
 
 		static auto legendMarker(const std::string_view &categoryName,
 		    const std::string_view &categoryValue,
-		    Gen::ChannelId channel)
+		    Gen::ChannelId channel,
+		    double yOverflow)
 		{
 			return std::make_unique<CategoryInfo<LegendChild>>(
 			    categoryName,
 			    categoryValue,
 			    "marker",
-			    channel);
+			    channel,
+			    yOverflow);
 		}
 
-		static auto legendBar(Gen::ChannelId channel)
+		static auto legendBar(Gen::ChannelId channel,
+		    double yOverflow)
 		{
-			return std::make_unique<LegendChild>("bar", channel);
+			return std::make_unique<LegendChild>("bar",
+			    channel,
+			    yOverflow);
 		}
 
 		static auto dimAxisLabel(const std::string_view &categoryName,
Index: src/chart/rendering/drawlegend.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/chart/rendering/drawlegend.h b/src/chart/rendering/drawlegend.h
--- a/src/chart/rendering/drawlegend.h	(revision 3d3fc0a53c69be54e1abb0e9dd71e5ea001a209e)
+++ b/src/chart/rendering/drawlegend.h	(date 1724677684853)
@@ -30,6 +30,7 @@
 		Gfx::ICanvas &canvas;
 		Geom::Rect titleRect;
 		Geom::Rect markerWindowRect;
+		double yOverflow{};
 		double yOffset{};
 		Gen::ChannelId type{};
 		double weight{};
```

src/apps/weblib/typeschema-api/styles.yaml
```diff
@@ -1,14 +1,21 @@
 ---
 definitions:
-    Length:
+    LengthUnit:
```
simzer, line 3: Erre miért van szükség? 
Általánosságban: a style-oknál a vezérelv az volt, hogy amennyire csak lehet, próbáljunk meg CSS kompatibilisek maradni. Ez, hogy bevezetjük a + operátort a length tipusú propertiknél túl mutat ezen.
Ha nem must have, én nem csinálnám. Ha must have, akkor CSS kompatibilisen a calc()-ról kellene mintázni: https://developer.mozilla.org/en-US/docs/Web/CSS/calc

schaumb: Én próbáltam a toString/fromString-jét a Gfx::Length-nek egységesíteni, mert a toString (operator std::string) így működik ('+' jellel elválaszva a 3 komponenst (%/px/em) kiírja).

ennek nem feltétlenül kell kifelé is látszania

simzer: Ah értem.
Akkor az bizti, hogy kifelé nem kellene látszania.
Kicsit kilóg nekem ebből az osztályból, hogy kifejezést értékel ki.

schaumb:
Add a changelog from this PR https://github.com/vizzuhq/vizzu-lib/pull/569

src/chart/rendering/drawlegend.h
```diff
@@ -25,20 +25,35 @@ class DrawLegend : public DrawingContext
 	    *rootStyle.plot.marker.colorGradient};
 
 private:
+	struct ColorGradientSetter
+	{
+		Geom::Line line;
+		Gfx::ColorGradient gradient;
```
simzer, line 31: Would be good to have a `Gfx::LinearGradient` type for line+gradient

src/chart/rendering/drawlegend.cpp
```diff
@@ -28,48 +32,97 @@ void DrawLegend::draw(Gfx::ICanvas &canvas,
...
 	    .dimension = plot->axises.at(channelType).dimension,
-	};
+	    .colorGradientSetter = {markerWindowRect.leftSide(),
+	        Gfx::ColorGradient{{
+	            {0.0, {}},
+	            {fadeElementPercent / 2.0, {}},
+	            {fadeElementPercent, {}},
+	            {1.0 - fadeElementPercent, {}},
+	            {1.0 - fadeElementPercent / 2.0, {}},
+	            {1.0, {}},
```
simzer, line 71: Why do we need 6 points here instead of three?

src/chart/rendering/drawlegend.cpp
```diff
@@ -28,48 +32,97 @@ void DrawLegend::draw(Gfx::ICanvas &canvas,
...
+	for (auto &stop : modifiableStops) stop.value = color;
+
+	modifiableStops[0].value.alpha = 0.0;
+	modifiableStops[1].value.alpha *= 0.27;
+	modifiableStops[4].value.alpha *= 0.27;
+	modifiableStops[5].value.alpha = 0.0;
+
+	canvas.setBrushGradient(
+	    {transform(line.begin), transform(line.end)},
+	    gradient);
```
simzer, line 117: I think this setter is a bit overcomplicated for the problem. 
What we would like to be able to do is to set a transparency gradient for the label, if needed.  
So why not simply give a `Math::SegmentedFunction<double> transparencyGradient, Geom::Line gradientPosition` parameters for `DrawLabel` and calculate the color gradient in DrawLabel from the calculated text color?

src/chart/rendering/drawlegend.cpp
simzer: If I understand correctly, this way we create an S curve like gradient function. The goal of the three point (0,0 .5,.27 1,1) was to approximate this curve https://en.wikipedia.org/wiki/Lightness#/media/File:Lightness_approximations.svg

src/chart/rendering/drawlegend.h
schaumb: but for this class, the "line" and the colors are changing always.

src/chart/rendering/drawlegend.cpp
simzer: On a second tought, we shouldn't even doing that, since background/foreground color could be changed, and for a dark theme we should use an inverse curve. So to be on the safe side, we should have a simple 2 point linear gradient

schaumb: These 6 points represent the whole fade area (upper 3 and lower 3 too) and handle both in the same class.
This is because, between the fade areas, the markers need to be fully colored, and I didn't want to handle differently the 2 sides.

src/apps/weblib/ts-api/events.ts
```diff
@@ -124,6 +124,8 @@ export interface Caption extends TextElement {
 export interface Legend extends Element {
 	tagName: 'legend'
 	channel: string
+	/** Marker labels y overflow by pixel */
+	yOverflow: number
```
simzer, line 128: Would be good to have a HTML element compatible parameter here, since the events and their elements are mimicking DOM events and Elements. 
There is no analogous parameter in DOM elements, the closest which used for the same purpose: 
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight

src/chart/rendering/drawlegend.cpp
schaumb: because it's not just the label but the marker needs to. And that will be a duplication for the same gradient problem.

src/apps/weblib/ts-api/events.ts
schaumb: I'll rename `yOverflow` to `scrollTop`.

src/chart/rendering/drawlegend.cpp
schaumb: So the 

```
egy egyszerű gamma-corrected közelítés:
0% - alpha=1.0
50% - alpha=0,27
100% - alpha=0.0
```
message is irrelevant?

schaumb: OK. I'll remove the 0.27 points.

src/chart/rendering/drawlegend.cpp
simzer: I see. It's unfortunate. Still could be more clean. I see three problems:
-  This class encapsulates three things (building of the ColorGradient, transforming the line, and setting it to the canvas). I think it would be better if it would have a singular purpose, building the ColorGradient.
- How we build the color gradient is that we have an alpha/transparency gradient, and we get the color from outside.
  the class does not represent this, since it contains a colorgradient, and the transparency gradient is hard coded in the function.
  I think a clearer and more generic approach would be to have a class, which stores an arbitrary alpha gradient and has a method which accepts a color and builds the ColorGradient.
  
Something like this:
  
 ```cpp
 /// legend draw:
 fadeoutLine = markerWindowRect.leftSide()
 fadeoutGradient = AlphaGradient({ 
   {0.0, 0.0}, 
   {fadeElementPercent, 1.0}, 
   {1.0 - fadeElementPercent, 1.0}, 
   {1.0, 0.0} 
 })
 
 // marker draw and label draw
 canvas.setBrushGradient(transform(fadeoutLine), fadeoutGradient.toColorGradient(color))
 ```

simzer: Also, draw label is a more generic lower level utility class then legend draw. I think it's not a good practice to make the interface of the lower level class more cryptic to optimise code duplication in the higher level user class.

src/apps/weblib/ts-api/events.ts
simzer: So, the style parameters are named based on CSS attributes, and Element properties named based on HTML element property names.
The equivalent would be in JS:
```
LegendElement.style.translateY = xxx
console.log(LegendElement.scrollTop)
```
So translateY is ok, because it is CSS like enough. But in the event element we need HTMLElement-like names.

Also scrollTopMax is non standard and it does not give enough information, since it does not represent the actual position, but the max position.

schaumb: The actual position is represented by the style parameter which will be redundant if the event contains too

simzer: The style parameter is showing the target position, not the actual position during animation.

src/base/math/segmentedfunc.tpp
```diff
@@ -1,73 +1,31 @@
...
-template <typename T>
-T Math::SegmentedFunction<T>::at(double pos) const
+template <typename T, class CRTP>
+T Math::SegmentedFunction<T, CRTP>::operator()(double pos) const
 {
 	if (stops.empty()) return T();
-	if (stops.size() == 1) return stops.at(0).value;
-
-	if (pos < stops.front().pos) return stops.front().value;
-	if (pos > stops.back().pos) return stops.back().value;
```
simzer, line 16: This was an optimization, to prevent iterating over the stop array. This usecase of having points outside the range is frequently occurs in animation easings

simzer (approved):

src/base/math/segmentedfunc.tpp
schaumb: I revert this `if` here: https://github.com/vizzuhq/vizzu-lib/pull/570/files#diff-6ec5da3cb41af87a90a576b0263ae1367eb23cdffb3770aca583dd4f2a62feefR76

PR #570, opened 2024-08-30
Remove "*this = " expressions from codebase

schaumb:


src/base/text/funcstring.cpp
```diff
@@ -22,15 +23,18 @@ FuncString::FuncString(std::string code, bool throwOnError)
 	}
 
 	if (parts[1].empty() || parts[1].back() != ')') {
-		if (!throwOnError) return;
+		if (!throwOnError) {
```
simzer, line 26: please check if changed functionality does not cause problem where it is used

schaumb: I reverted this modification

simzer (approved):

PR #572, opened 2024-09-04
Remove text bg + Fix alphas

schaumb:


src/base/gfx/colortransform.cpp
```diff
@@ -73,6 +73,15 @@ ColorTransform ColorTransform::Opacity(double factor)
 	    "opacity(" + std::to_string(factor) + ")"};
 }
 
+ColorTransform ColorTransform::Fade(double factor)
```
simzer, line 76: We should call this opacity, because this is what CSS opacity does. What is now opacity directly sets the alpha channel, so we should call that alpha instead.

src/base/math/fuzzybool.h
```diff
@@ -124,6 +124,11 @@ class FuzzyBool
 		return FuzzyBool(std::max(0.0, 1 - 2 * value));
 	}
 
+	[[nodiscard]] FuzzyBool lessAndMore() const
```
simzer, line 127: It seems more like moreOrLess()
```
more() || less() 
== max(more(), x.less())
== max( max(0.0, 2 * value - 1), max(0.0, 1 - 2 * value) )  
== value < 0.5 ? max(1 - 2 * value) : max(0.0, 2 * value - 1)
== value < 0.5 ? 1 - 2 * value : 2 * value - 1 
== asb( 2 * value - 1 )
```

src/base/math/fuzzybool.h
```diff
@@ -134,6 +139,43 @@ class FuzzyBool
...
+	{};
+
+	template <class Res, class... Args>
+	using CommonOrFuzzy = std::conditional_t<std::is_void_v<Res>,
+	    typename CommonOrFuzzyImpl<std::common_type<Args...>>::type,
+	    Res>;
+
+	template <class Res = void, class Arg, class... Args>
+	[[nodiscard]] static CommonOrFuzzy<Res, Arg, Args...>
+	And(Arg &&arg, Args &&...args)
```
simzer, line 158: I'm not following. why is this `And(a,b,c) ` better then `a && b && c`

src/chart/rendering/drawlabel.cpp
```diff
@@ -63,15 +49,12 @@ void DrawLabel::draw(Gfx::ICanvas &canvas,
...
-	if (options.alpha) {
-		if (auto textColor = *style.color * *options.alpha;
-		    options.gradient)
-			options.gradient(canvas,
-			    textTransform.inverse(),
-			    textColor);
-		else
-			canvas.setBrushColor(textColor);
-	}
+	if (auto &&textColor = options.colorTransform(*style.color);
```
simzer, line 52: Since we removed the background, i'm not sure if it still makes sense to setup the brush here. We could setup it outside, then call DrawLabel

simzer (approved):

PR #579, opened 2024-09-13
On treemap switch off rectangleSpacing

schaumb:


schaumb:
This PR is not acceptable. The rectangleSpacing option is affecting only the dimension channels. Treemap is a measure (size) channel.

PR #586, opened 2024-10-03
Remove align min

schaumb:


src/chart/options/align.h
```diff
@@ -10,15 +10,14 @@ namespace Vizzu::Base
...
+	enum class Type : std::uint8_t { none, center, stretch };
 
 	Align(Type type, const Range &range) : Range(range), type(type) {}
 
 	[[nodiscard]] Range getAligned(const Range &range) const
 	{
 		switch (type) {
-		case Type::none: return range;
-		case Type::min: return range + min - range.getMin();
+		case Type::none: return range - range.getMin() + min;
```
simzer, line 20: I think at none we should just return the range unchanged as before. otherwise it will do what min did before, which was the wrong behavior on negative values. 
Consider this chart, where it works incorrectly with none:
https://jsfiddle.net/simzer/rj4hc0m2/2/
We show the negative value as positive. 

It is not causing problem in PlotBuilder::addAlignment only because we return from that function if align is none before we would apply the alignment. This is contraintuitive.

simzer (changes requested):

simzer (approved):

src/chart/options/align.h
schaumb: Only used at splited charts, this will be removed later

simzer (approved):

PR #599, opened 2024-11-04
Axis refactor v5 - Remove MarkerPosition, move to common creation

schaumb:


src/dataframe/impl/data_source.cpp
```diff
@@ -533,7 +562,7 @@ std::string data_source::get_id(std::size_t record,
 		const auto *val = dimensions[ix].get(record);
 		res += name;
 		res += '\37';
-		res += val ? *val : std::string_view{"␀"};
+		res += val ? *val : std::string_view{"\33"};
```
simzer, line 565: We should clean chars lower then 0x20 just to be safe

simzer (approved):

PR #601, opened 2024-11-05
API: Add array begin/end

schaumb:


src/chart/options/channel.h
```diff
@@ -71,14 +71,18 @@ struct ChannelSeriesList
 
 	[[nodiscard]] std::string toString() const;
 
+	struct RawChannel
```
simzer, line 74: RawChannelSeries.

simzer (approved):

PR #602, opened 2024-11-06
Axis refactor v7 - Legend interpolation, Try to separate axis generator functions

schaumb:


src/chart/generator/axis.h
```diff
@@ -53,13 +55,17 @@ struct MeasureAxis
 {
 	::Anim::Interpolated<bool> enabled{false};
 	Math::Range<double> range = Math::Range<double>::Raw(0, 1);
+	std::string series;
 	::Anim::String unit;
```
schaumb, line 59: Unit not anim, axis not needed

src/chart/generator/axis.h
```diff
@@ -81,17 +87,22 @@ struct DimensionAxis
 		Math::Range<double> range;
 		double value;
 		::Anim::Interpolated<ColorBase> colorBase;
-		::Anim::String label;
-		std::string categoryValue;
+		::Anim::Interpolated<bool> label;
```
schaumb, line 90: Not interpolated, but removed if no label needed.

src/chart/generator/axis.h
schaumb: Unit is anim. (%)

src/chart/generator/axis.h
schaumb: not remove the labels, because interpolation is useful anyway.

simzer (approved):

PR #613, opened 2024-11-17
Axis refactor v12 - Separate AxisChannel properties from Channel properties 

schaumb:


src/apps/weblib/typeschema-api/config.yaml
```diff
@@ -76,6 +76,12 @@ definitions:
                 oneOf:
                     - type: number
                     - { type: string, enum: [auto] }
+
+    AxisChannel:
+        description: |
```
simzer, line 81: Channel props can be inherited:
```yaml
        $extends: Channel
```

src/apps/weblib/typeschema-api/config.yaml
```diff
@@ -109,14 +115,18 @@ definitions:
         type: object
         properties:
             x:
-                $ref: Channel
+                allOf:
```
simzer, line 118: no need for `allOf` if `AxisChannel` inherits `Channel` properties

src/chart/options/channel.h
```diff
@@ -118,9 +118,22 @@ struct ChannelSeriesList
 	bool removeSeries(const Data::SeriesIndex &index);
 };
 
-class Channel
+struct AxisChannelProperties
```
simzer, line 121: Wouldn't it make sense if instead of `AxisChannelProperties`, we had an `AxisChannel` derived from `Channel`?

src/chart/options/channels.h
```diff
@@ -15,6 +15,7 @@ namespace Vizzu::Gen
 struct Channels : Refl::EnumArray<ChannelId, Channel>
 {
 	using IndexSet = std::set<Data::SeriesIndex>;
+	EnumArray<AxisId, AxisChannelProperties> axisProps;
```
simzer, line 18: Wouldn't something like this would be more straight forward:
```
struct Channels
{
	Refl::EnumArray<NonAxisId, Channel> nonAxisChannels;
	EnumArray<AxisId, AxisChannel /* inherits Channel */> axisChannels;

src/chart/options/channel.h
schaumb: The reflection system requires the creation of another class for that.

src/chart/options/channels.h
schaumb: Yes. Later I'll fix that.

src/apps/weblib/typeschema-api/config.yaml
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/632/commits/bbbc5238e656e889b13c997566ff6cb833ae195d

src/apps/weblib/typeschema-api/config.yaml
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/632/commits/bbbc5238e656e889b13c997566ff6cb833ae195d

src/chart/options/channels.h
schaumb: https://github.com/vizzuhq/vizzu-lib/pull/632/commits/bbbc5238e656e889b13c997566ff6cb833ae195d

simzer (approved):

PR #617, opened 2024-11-25
Axis refactor v11e - Cross interlacing, NaN handling, off dimLabel out

schaumb:


src/chart/rendering/drawinterlacing.cpp
```diff
@@ -34,6 +34,37 @@ void DrawInterlacing::drawGeometries(Gen::AxisId axisIndex) const
 	    || guides.interlacings == false)
 		return;
 
+	std::map<double, double> othWeights{{0.0, 0.0}, {1.0, 0.0}};
```
simzer, line 37: `othWeights` i think it's better to not leave off two characters, `otherWeights` would be more readable. Same goes for Guides and AxisStyle

src/chart/rendering/drawinterlacing.cpp
```diff
@@ -48,30 +79,64 @@ void DrawInterlacing::drawGeometries(Gen::AxisId axisIndex) const
...
+					            std::abs(
+					                axisStyle.interlacing.color->blue
+					                - color.blue)}),
+					    color,
+					    2.0 - (weight + first->second));
+				drawInterlacing(first->second > weight ? !axisIndex
+				                                       : axisIndex,
+				    color,
+				    rect(first->first, next->first));
+			}
```
simzer, line 138: this function got too big, I think it would help readability to separate it to well named functions.

simzer (approved):

PR #618, opened 2024-11-25
Axis refactor v11f - Add split style parameter for plot.axes

schaumb:


src/chart/options/channel.h
```diff
@@ -63,7 +63,7 @@ static_assert(std::ranges::all_of(Refl::enum_names<LegendId>,
 	            Refl::get_enum<ChannelId>(name));
     }));
 
-[[nodiscard]] constexpr Geom::Orientation operator+(
+[[nodiscard]] constexpr Geom::Orientation operator++(
```
simzer, line 66: `++` still not intuitive, I think it make sence to convert between Axis and Channel ID with an operator, since they are the same set of values, only that Axis is a subset.
However Orientation and AxisID are different, I would use a conversion function instead, IMHO it would make the code more readable.

src/chart/main/style.h
```diff
@@ -245,6 +245,7 @@ struct Axis
 	Tick ticks;
 	Guide guides;
 	Interlacing interlacing;
+	Param<Gfx::Length> split;
```
simzer, line 248: `split` is not really describing the parameter well. What we are setting here is the length of the spacing when the chart is split. I think `spacing` would be a better name. It would also be closer to CSS naming convention, what we try to follow in the styles.

src/apps/weblib/typeschema-api/styles.yaml
```diff
@@ -452,6 +452,9 @@ definitions:
             interlacing:
                 $ref: Interlacing
                 nullable: true
+            split:
+                $ref: Length
```
simzer, line 456: If the parameter can only be set in percentage, then the type should reflect that.
`{ type: string, mask: /:number:%/ }` instead of `Length`

simzer (dismissed):

simzer (approved):

